PR #13, opened 2022-06-17
Fix engines dependency

flecoufle:
minimum requirement engines is >= 14
but in coherence with the platfrom >=16


hardillb (approved):

hardillb:
Thanks for this, it makes sense

PR #22, opened 2022-07-22
Add MQTT connectivity support

knolleary:
This PR adds MQTT support to the device agent.

In the old code, the agent consisted of a Pinger class that was responsible for regularly checking in with the platform to see what it should be doing. This class kept the process alive.

This PR has refactored the structure of the device agent:

 - `Agent` - this is the heart of the agent. It has a function that can be used to update the state of the agent (ie project/snapshot/settings updates).
 - The agent then creates an `HTTPClient` used to retrieve snapshots/settings when needed.
 - If an MQTT configuration is provided, it also creates an `MQTTClient` object that does the necessary connecting and subscribing.
 - If no MQTT config is provided, the `HTTPClient` is told to start polling the platform for updates
 - When either the HTTPClient or MQTTClient determines a change in config is needed, they notify the Agent.

This model makes it easier to test the core update logic of the Agent without needing to setup test brokers or http servers.

This PR requires https://github.com/flowforge/flowforge/pull/706 on the platform side.

To enable the MQTT feature, the following options must be added to `device.yml`:

MQTT options   | Description
---------------|---------------
`brokerURL`      | The url for the platform broker
`brokerUsername` | The username to connect with - `device:<teamId>:<deviceId>`
`brokerPassword` | The password to connect with


knolleary:
This is good for reviewing.

Have tested against staging deploy - all working as expected.

hardillb (approved):

PR #28, opened 2022-08-12
Clean up .config.nodes.json

### LINKED ISSUE #27: Changing project on device doesn't remove old modules
### Current Behavior

When I change the project that my device is running and deploy a new snapshot for that project Node-RED still picks up and tries to install modules from the previous project.
I think this is due to `/opt/flowforge-device/project/node_modules` not being wiped.

### Expected Behavior

Deploying a new snapshot should result in a clean install of only the modules in the snapshpot

### Steps To Reproduce

Build project with a custom node.
Deploy to device,

Build new project without custom node.
Deploy to device.

Original custom node is still started.

### Environment

- FlowForge version: 0.8.0
- Agent 0.2.0
- Node.js version: 16.16
- npm version:
- Platform/OS:
- Browser:

hardillb:
If we don't remove these files then Node-RED will try and reinstall old nodes when snapshot changes

fixes #27

### DIFF
--- a/lib/launcher.js
+++ b/lib/launcher.js
@@ -43,6 +43,8 @@ class Launcher {
         const packageJSON = JSON.parse(JSON.stringify(packageJSONTemplate))
         packageJSON.dependencies = JSON.parse(JSON.stringify(this.snapshot.modules))
         await fs.writeFile(this.files.packageJSON, JSON.stringify(packageJSON, ' ', 2))
+        await fs.unlink(path.join(this.projectDir, '.config.nodes.json'))
+        await fs.unlink(path.join(this.projectDir, '.config.nodes.json.backup'))
     }
 
     async installDependencies () {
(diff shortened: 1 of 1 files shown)


lib/launcher.js
```diff
@@ -43,6 +43,8 @@ class Launcher {
         const packageJSON = JSON.parse(JSON.stringify(packageJSONTemplate))
         packageJSON.dependencies = JSON.parse(JSON.stringify(this.snapshot.modules))
         await fs.writeFile(this.files.packageJSON, JSON.stringify(packageJSON, ' ', 2))
+        await fs.unlink(path.join(this.projectDir, '.config.nodes.json'))
+        await fs.unlink(path.join(this.projectDir, '.config.nodes.json.backup'))
```
knolleary, line 47: May be better to use `fs.rm( path, { recursive: true } )` so it copes with them files not existing.

hardillb: Update pushed, I think it should be `{ force: true }` not `{ recursive: true }`

sammachin (approved):

PR #35, opened 2022-10-13
add jitter to mqtt and http checkin

### LINKED ISSUE #23: Add jitter to timings to avoid connection/publish storms
There are a number of actions the agent takes in response to events that cause it to ping the platform. For a single device that's fine. But at scale, that can cause problems.

We should introduce some variance to those timings to avoid the platform being overloaded by requests.

Scenarios that need considering:

 - When a new active snapshot is set, the platform publishes to all devices. In response, all devices make an HTTP request to get the new snapshot.
 - When the device successfully connects to the broker, it publishes an initial check-in
 - ... there may be more...
 - The regular heartbeat ping (both mqtt and http) should include some variance - eg +/- 10 seconds

Steve-Mcl:
fixes #23 

NOTES...

A new local class `IntervalJitter` was made to keep the logic changes in the main src to a minimum. By that I mean something capable of having an initial (lower) jittered delay (for 1st checkin) and then stable jitter centered around the desired "interval" +- jitter
 
Introducing jitter was surprisingly more difficult than I first envisioned - nothing in NPM suited my needs.  The majority of examples use this **nested timeout** pattern...
```js
    start () {
        this.stopped = false
        const runner = () => {
            // Stop if this jitter was stopped.
            if (this.stopped) { return }
            const interval = addJitter(this.#min, this.#max)
            this.intervalTimer = setTimeout(() => {
                if (this.stopped) { return }
                this.#callback && this.#callback(this)
                this.#loopTimer = setTimeout(runner, 1)  // << **** nested timeout ****
            }, interval)
        }
        runner()
    }
``` 

Where every call ends up adding to the stack and over the course of time, the delay is never centered around the set "interval" delay.  In tests, most aproaches around the web and in npm end up miles off course.

The included solution uses a unique pattern to avoid this.

For example, the below code ...
```js
const j1 = new IntervalJitter()
const timings = []
let first = true
j1.on('started', () => {
    console.log('started')
})
j1.on('interval', (t) => {
    if(!first) { timings.push(t) }
    console.log(`j1 @ ${t}ms`)
    first = false 
})
j1.on('stopped', () => {
    console.log('stopped')
    console.log('Minimum delay :', Math.min(...timings))
    console.log('Maximum delay :', Math.max(...timings))
    console.log('Average delay :', timings.reduce( ( a, b ) => a + b, 0 ) / timings.length)
    console.log('Total delay   :', timings.reduce((a, b) => a + b, 0))
})
j1.start({interval: 1000, jitter: 200, firstInterval: 10, firstJitter: 100})
setTimeout(() => {  j1.stop() }, 10100)
```

results in a jitter centered firmly around the desired "interval" **without** filling the stack with nested `setTimeouts`...
```log
started
j1 @ 18ms    << first delay is between 10 ~ 100ms as required
j1 @ 1007ms
j1 @ 1119ms
j1 @ 1062ms
j1 @ 1001ms
j1 @ 965ms
j1 @ 1033ms
j1 @ 920ms
j1 @ 1058ms
j1 @ 965ms
j1 @ 887ms   << subsequent delays all hover around 1000ms as requested
stopped
Minimum delay : 887ms
Maximum delay : 1119ms
Average delay : 1001.7ms
Total delay   : 10017ms
```

And the result of `interval: 50`, `jitter: 50` over 10 secs ...

```log
Minimum delay : 3
Maximum delay : 97
Average delay : 50.43434343434343
Total delay   : 9986
```

### DIFF
--- a/lib/IntervalJitter.js
+++ b/lib/IntervalJitter.js
@@ -1,0 +1,99 @@
+const EventEmitter = require('events');
+/**
+ * An interval timer with jitter
+ * @example
+ * const j1 = new IntervalJitter()
+ * j1.start({interval: 1000, jitter: 100}, (t) => { console.log(`hello j1, this was delayed ${t}ms`) })
+ * setTimeout(() => { j1.stop() }, 5000)
+ * @example
+ * const j2 = new IntervalJitter()
+ * j2.on('interval', (t) => { console.log(`hello j2, this was delayed ${t}ms`) })
+ * j2.on('started', () => { console.log('j2 started') })
+ * j2.on('stopped', () => { console.log('j2 stopped') })
+ * // Start with initial "interval" firing at somewhere between 0 ~ 500ms
+ * // After that, fire somewhere between 900 ~ 1100ms
+ * j2.start({interval: 1000, jitter: 200, firstInterval: 0, firstJitter: 500})
+ * setTimeout(() => { j2.stop() }, 5000)
+ */
+class IntervalJitter extends EventEmitter {
+    constructor () {
+        super()
+        this.interval = 1000
+        this.jitter = 1000
+        this.firstInterval = null
+        this.firstJitter = null
+        this.callback = null
+        this.#init()
+    }
+    #stopped = true
+    #initTimer
+    #intervalTimer
+    #jitterTimer
+    #init() {
+        const self = this
+        self.on('internal:start1', (interval, jitter) => {
+            self.emit('started')
+            let firstDelay = interval + IntervalJitter.calculateJitter(jitter)
+            self.#initTimer = setTimeout(() => {
+                if (self.#stopped) { return }
+                self.callback && self.callback.call(this, firstDelay)
+                self.emit('interval', firstDelay)
+                self.lastTime = Date.now()
+                self.emit('internal:start2', self.interval, self.jitter)
+            }, firstDelay)
+        })
+        self.on('internal:start2', (interval, jitter) => {
+            if (self.#stopped) { return }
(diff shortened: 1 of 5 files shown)


knolleary:
Can you update the readme with details of the default jitter behaviour and how it can be configured?

Steve-Mcl:
> Can you update the readme with details of the default jitter behaviour and how it can be configured?

Hi Nick, do you mean the parent issue #23 or some other readme?

knolleary:
I mean the repo readme: https://github.com/flowforge/flowforge-device-agent/blob/main/README.md

If the jitter adds a new config option, it needs to be documented.

Steve-Mcl:
> Can you update the readme with details of the default jitter behaviour and how it can be configured?

Resolved in [cc403fb](https://github.com/flowforge/flowforge-device-agent/pull/35/commits/cc403fb4710a924ef477bb5636c8f76ae493879b)

knolleary (approved):

lib/IntervalJitter.js
```diff
@@ -0,0 +1,99 @@
...
+    constructor () {
+        super()
+        this.interval = 1000
+        this.jitter = 1000
+        this.firstInterval = null
+        this.firstJitter = null
+        this.callback = null
+        this.#init()
+    }
+    #stopped = true
```
knolleary, line 28: If you are going to use private class members you'll need to update `ecmaVersion` to `latest` in `.eslintrc`

knolleary (changes requested):

knolleary (approved):

PR #44, opened 2022-12-08
Add "memory" and "persistent" as context stores

Steve-Mcl:
part of flowforge/flowforge#212

Adds "memory" and "persistent" context storage entries to the device as agreed

Please update checklist here: https://github.com/flowforge/flowforge/issues/212#issuecomment-1342852768 once complete

NOTE: Implementing this here (as opposed to the settings.json generated for a device by forge) means the device agent must be updated before the context stores are realised.

### DIFF
--- a/lib/template/template-settings.js
+++ b/lib/template/template-settings.js
@@ -16,5 +16,10 @@ module.exports = {
         }
     },
     credentialSecret: settings.credentialSecret,
-    flowforge: settings.flowforge
+    flowforge: settings.flowforge,
+    contextStorage: {
+        default: 'memory',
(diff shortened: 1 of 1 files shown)


knolleary (approved):

knolleary:
> NOTE: Implementing this here (as opposed to the settings.json generated for a device by forge) means the device agent must be updated before the context stores are realised.

I think it is true to say that even if the context options were to be passed to the device agent by the platform, this file would still need to be updated so as to pick them up anyway.

Given we don't have any plans that involve being able to dynamically change the context stores with this feature, I support this approach of hardcoding in this configuration.

Issue #45, opened 2022-12-21
Add flag to skip the `npm install` step of deploying snapshot


knolleary:
### Description

**As a:** user on a secure network

**I want to:** run the device agent without it running `npm install` each time a new snapshot is deployed

**So that:** I can use the agent on an air-gapped network.

---

The device agent current runs `npm install` each time a new snapshot is deployed. This is done to ensure Node-RED and all dependencies are installed at the right level.

If the device is running on an air-gapped network, without any access to the internet or npm registry, then this won't work.

However, if the user has ensured Node-RED and all required npm modules are preinstalled on the device, then the device agent should still be able to run.

The proposal here is to add a flag to the device agent configuration to skip the npm install step.

---

Implemented via #47, we have added a new command-line parameter to the device agent (`-m / --moduleCache`) that tells it to skip running `npm install` when loading a snapshot and instead expects a local directory (`module_cache/node_modules`) to exist that contains all of the require modules preinstalled.

How to create the module cache is detailed in the doc updates included in that PR.

### Acceptance Criteria

 - Can run the device agent without it having to npm install
 - Pre-installed nodes can be used by the user
 - Docs needed for how to use this in practise

knolleary:
Resolved by #47

PR #47, opened 2023-01-05
Support running on offline devices

hardillb:
## Description

This allows Node-RED and nodes to be loaded from a local cache rather than being installed from npmjs.org each time the project snapshot is updated.

## Related Issue(s)

part of #45

## Checklist

    - [ ] Upgrade instructions
    - [x] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,21 +111,26 @@ Global Options
 By default the Device Agent will try and download the correct version of Node-RED and 
 any nodes required to run the Project Snapshot that is assigned to run on the device.
 
-If the device is being run on a offline network or security policies prevent the 
+If the device is being run on an offline network or security policies prevent the 
 Device Agent from connecting to npmjs.org then it can be configured to use a pre-cached 
 set of modules.
 
 You can enable this mode by adding `-m` to the command line adding `moduleCache: true` 
 to the `device.yml` file. This will cause the Device Agent to load the modules from the 
-`module_cache` directory in the Device Agents Configuration directory as describe above.
+`module_cache` directory in the Device Agents Configuration directory as described above.
 By default this will be `/opt/flowforge-device/module_cache`.
 
-The easiest way to create the cache is to download the `package.json` for the Snapshot. 
-This can be found in the 3 dots menu for the Snapshot on the Project's Snapshot page.
+### Creating a module cache
 
-Place this file in an empty directory on a machine with the same OS and architecture as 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -104,6 +106,27 @@ Global Options
...
+to the `device.yml` file. This will cause the Device Agent to load the modules from the 
+`module_cache` directory in the Device Agents Configuration directory as describe above.
+By default this will be `/opt/flowforge-device/module_cache`.
+
+The easiest way to create the cache is to download the `package.json` for the Snapshot. 
+This can be found in the 3 dots menu for the Snapshot on the Project's Snapshot page.
+
+Place this file in an empty directory on a machine with the same OS and architecture as 
+the device and run `npm install`. This will create a `node_modules` directory which you 
+should copy into `module_cache` director on the device.
```
knolleary, line 128:
```suggestion
By default the Device Agent will try and download the correct version of Node-RED and 
any nodes required to run the Project Snapshot that is assigned to run on the device.

If the device is being run on an offline network or security policies prevent the 
Device Agent from connecting to npmjs.org then it can be configured to use a pre-cached 
set of modules.

You can enable this mode by adding `-m` to the command line adding `moduleCache: true` 
to the `device.yml` file. This will cause the Device Agent to load the modules from the 
`module_cache` directory in the Device Agents Configuration directory as described above.
By default this will be `/opt/flowforge-device/module_cache`.

### Creating a module cache

To create a suitable module cache, you will need to install the modules on a local device with
access to npmjs.org, ensuring you use the same OS and Architecture as your target
device, and then copy the modules on to your device.

1. From the Project Snapshot page, select the snapshot you want to deploy and select the option to download its `package.json` file.
2. Place this file in an empty directory on your local device.
3. Run `npm install` to install the modules. This will create a `node_modules` directory.
4. On your target device, create a directory called `module_cache` inside the Device Agent Configuration directory.
5. Copy the `node_modules` directory from your local device to the target device so that it is under the `module_cache` directory.
```

A slight rewording and given the steps to create the module cache some more explicit structure.

knolleary (changes requested):
Some docs rewording.

As this directly references the ability to download `package.json`, I'd like to get that PR reviewed and merged first.... next on my list to look at.

hardillb:
Tthere is exactly the same wording in the doc update for the flowoforge project as well

knolleary (approved):

Issue #53, opened 2023-01-20
Facing SyntaxError: Unexpected token  when I install and boot for the first time


hfc01512:
### Current Behavior

I was introduced to this software by an acquaintance and started using it.
I have stored the device file device.yml in place and when I try to start it, I get the following error message and cannot start it.
I would appreciate any advice on how to correct this.

-----------------------------

/usr/local/lib/node_modules/@flowforge/flowforge-device-agent/lib/agent.js:47
                info(`Snapshot: ${this.currentSnapshot?.id || 'none'}`)
                                                       ^

SyntaxError: Unexpected token .
    at Module._compile (internal/modules/cjs/loader.js:723:23)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at Object.<anonymous> (/usr/local/lib/node_modules/@flowforge/flowforge-device-agent/index.js:3:19)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
------------------------------
Thank you

### Environment

- FlowForge version: 
- Node.js version: v10.19.0
- npm version: 6.14.4
- Platform/OS: Ubuntu 20.04
- Browser: chrome

knolleary:
Hi @hfc01512 

I can see you are using Node.js v10. The device agent requires Node.js 16 or later. You will need to update your version of node.js to use the agent.

hfc01512:
Hell knolleary,
Thanks for the advice.
I have successfully activated the system.

hardillb:
fixed by #55

PR #55, opened 2023-01-31
Move nodejs version check

### LINKED ISSUE #54: Add check on node.js version
### Story

_No response_

### Description

We should add an explicit node.js version check to the device agent so we can provide a more helpful message if trying to run on an older version of node.js.

This will have to be done right at the start of `index.js` - *before* we require any other modules, as they will hit syntax errors when trying to do the require if node is too old.

Ref: https://github.com/flowforge/flowforge-device-agent/issues/53#issue-1551098547

hardillb:
## Description

Need to test NodeJS version before using features only available in NodeJS 16.x or newer

## Related Issue(s)

fixes #54

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,8 +1,15 @@
 #!/usr/bin/env node
+
+const semver = require('semver')
+
+if (semver.lt(process.version, '16.0.0')) {
+    console.log('FlowForge Device Agent requires at least NodeJS v16')
+    process.exit(0)
+}
+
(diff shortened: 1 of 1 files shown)


hardillb:
also fixes #53

knolleary (approved):

PR #56, opened 2023-02-08
Auto Device Provisioning

Steve-Mcl:
### Implements provisioning capability

To achieve the downloading of credentials, reloading the config, re-initialising the agent etc an AgentManager class was added to coordinate these operations.

### Commit notes...

> Adds a new class AgentManager that coordinates the provisioning
>  In particular file checks, device creation, file updates (fs-extra),
>  anti-repeat (to avoid unlimited device creation),
>  stopping, restarting and reloading the agent

To clarify the "anti repeat" part. If the device is provisioned in the Forge DB (i.e. a device is created from the provisioning token) we MUST ensure the "spent" provisioning token can be deleted BEFORE we generate the new device. So great care was put into ensuring files are accessible/writeable/delete-able before the "POST" to generate occurs.

> Moves the configuration from  index.js to the AgentManager -
>  in order to implement hot agent reload after provisioning

> Moves the device.yml existence checks out of config.js to index.js
>  to support agent reloading

> Adds the token checks to config.js to determine provisioning mode

> Added SIGTERM and SIGQUIT and refactored exiting to
>  ensure cleanup and halting

> Adds each class to its own exports (for type checking and improved DX)

Tests were updated/added but in interest of keeping the review size down and getting it out sooner, I have only implemented very basic tests.

## Related Issue(s)

[Feature Request: Automatic device join to project or easy setup scritpt for mass deploys#1212](https://github.com/flowforge/flowforge/issues/1212)

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

 - [-] Backport needed? -> add the `backport` label
 - [-] Includes a DB migration? -> add the `area:migration` label

### DIFF
--- a/lib/AgentManager.js
+++ b/lib/AgentManager.js
@@ -139,36 +139,39 @@ class AgentManager {
     }
 
     async provisionDevice (provisioningURL, provisioningTeam, provisioningToken) {
+        debug('Provisioning device')
+
+        // sanity check the parameters
+        provisioningURL = provisioningURL || this.configuration.forgeURL
+        provisioningTeam = provisioningTeam || this.configuration.provisioningTeam
+        provisioningToken = provisioningToken || this.configuration.token
+        const provisioningConfig = {
+            provisioningMode: true,
+            provisioningTeam,
+            forgeURL: provisioningURL,
+            token: provisioningToken
+        }
+        let provisioningOK = false
+        let postResponse = null
+
         try {
-            debug('Provisioning device')
-
-            // sanity check the parameters
-            provisioningURL = provisioningURL || this.configuration.forgeURL
-            provisioningTeam = provisioningTeam || this.configuration.provisioningTeam
-            provisioningToken = provisioningToken || this.configuration.token
-            const provisioningConfig = {
-                provisioningMode: true,
-                provisioningTeam,
-                forgeURL: provisioningURL,
-                token: provisioningToken
-            }
-
             // before we do anything, check if the device can be provisioned
             // These checks will ensure files are writable and the necessary settings are present
             if (await this.canBeProvisioned(provisioningConfig) !== true) {
-                warn('Device cannot be provisioned. Check the logs for more information.')
-                this.reloadAgent(6000 /* 60000 */) // reload as the device.yml file may have been fixed or updated
-                return false
+                throw new Error('Device cannot be provisioned. Check the logs for more information.')
             }
 
             // Get the local IP address / MAC / Hostname of the device for use in naming
-            const { host, ip, mac } = await this._getDeviceInfo(provisioningConfig.forgeURL)
-            const type = 'Auto Provisioned Device'
+            const { host, ip, mac, forgeOk } = await this._getDeviceInfo(provisioningConfig.forgeURL)
+            if (!forgeOk) {
+                throw new Error('Unable to connect to the Forge Platform')
+            }
+            const type = 'Auto Provisioned' + (provisioningConfig.provisioningName ? ` [${provisioningConfig.provisioningName}]` : '')
             const team = provisioningConfig.provisioningTeam
-            const name = (host || ip) + (mac ? ` (${mac})` : '')
+            const name = ((host || ip) + (mac ? ` (${mac})` : '')) || 'New Device'
 
             // Provision this device in the forge platform and receive the device ID, credentials and other details
-            const response = await got.post(`${provisioningConfig.forgeURL}/api/v1/devices`, {
+            postResponse = await got.post(`${provisioningConfig.forgeURL}/api/v1/devices`, {
                 headers: {
                     'user-agent': `FlowForge Device Agent v${this.configuration?.version || ' unknown'}`,
                     authorization: `Bearer ${provisioningConfig.token}`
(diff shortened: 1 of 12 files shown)


lib/AgentManager.js
```diff
@@ -0,0 +1,414 @@
+// AgentManager: A class that manages the device agent.
+// It can start/stop and reload the agent
+
+const { initLogger, info, warn, debug } = require('./log')
+const { default: got } = require('got/dist/source')
+const agent = require('./agent')
+const os = require('os')
+const path = require('path')
+const fsExtra = require('fs-extra')
```
knolleary, line 9: What operation requires `fs-extra`? I find Node 16+ core `fs` has pretty much everything I ever resorted to `fs-extra` for.

knolleary:
Lots of test failures that need investigating.

lib/AgentManager.js
Steve-Mcl: Using async / promises, I could devise no simple means of determining if a file exists (depreciated in fs) and 

Happy to revise is necessary.

Steve-Mcl:
Will run this on linux to see why tests fail (they run on win box)

lib/AgentManager.js
knolleary: `fs.existsSync` works just fine and isn't deprecated. Require it from the non-promise fs module...

```
const { existsSync } = require ('fs')
```

Steve-Mcl:
@knolleary  Test sandbox was not being restored. Instead of directly setting the new agent to a stub, I wrapped it in a sinon.stub so that `sinon.restore` actually restored the stub. This then fixed the many failing tests.

 ready for re-review.

lib/AgentManager.js
Steve-Mcl: I'm not getting notifications of comments! Will revise now.

Steve-Mcl:
@knolleary worked out a few kinks with using fs and improved resilience of the agent when conditions are not perfect (e.g. forge offline would crash agent, certain file permission/RO conditions would cause the download & writing of token not possible)

Ready for review once more.

Apologies for the delays but I am much happier with it now.

knolleary:
>Ready for review once more.
> Apologies for the delays but I am much happier with it now.

As mentioned throughout this week, I'm out of the office next week and I had a hard stop of 5pm today. This has come in too late for me to look - it's just too big a change and needs manual verification and testing that I'm not in a position to do now. 

@hardillb please can you look at this on Monday?

Steve-Mcl:
> As mentioned throughout this week, I'm out of the office next week and I had a hard stop of 5pm today. This has come in too late for me


@knolleary sorry about that, as I mentioned in my reply, I didnt get any notifications yesterday about your request to remove fsextra (but I appreciate your desire to minimise imports).  Once I spotted the reply (around 3:30pm today) I did attempt to get this resolved before 5PM (and almost did) but I was not satisfied with it as a stable working finished article & knowing the risks of generating unlimited devices (should the file system be in an undesirable state for auto provisioning to be successful), it took me another hour of testing to get it to a place I was happy to publish.  I needed (in my own mind) to be happy with it before pushing to gh (testing device.yml as a permissions locked file/read-only/busy/locked by a process, on both Linux & windows etc etc etc) so took more time that I had hoped. I just felt it was too important to ship without thorough testing since I could/should not expect a reviewer to go to the same lengths as I did to ensure as many scenarios as possible were covered. I am very conscientious you know ;) 

What I fear may not be apparent in this particular instance is the many considerations made with regards to the overall device provisioning activity in mind (for example, before this PR, the Agent would crash should the API be offline or should the device file be locked or read-only or have a permissions issue there was the possibility of generating unlimited devices).  As an integrator, I would not expect a remote agent to crash simply because the API was offline (the effort to reboot xxx amount of devices is not a happy prospect). Additionally, as a FF customer, I would not expect an integrators error with file permissions to cause many multiple devices to be generated.  All of this testing and iteration took more time than I had hoped but I do hope (from a quality POV) it was worth the effort and delay. 

I hope this (rather long) list of excuses help you appreciate my lateness Nick 🤞

lib/AgentManager.js
Steve-Mcl: Resolved in [1fefb9f](https://github.com/flowforge/flowforge-device-agent/pull/56/commits/1fefb9f1eb71469e62c17b4f851a6705fcc56cf1)

hardillb (changes requested):
Just 2 small changes.

And I also have a question about the hostname/mac address if the device agent starts in a docker container. I will try and find a way to test that in a moment

lib/mqtt.js
```diff
@@ -4,7 +4,13 @@ const { IntervalJitter } = require('./IntervalJitter')
 const url = require('url')
 
 class MQTTClient {
+    /**
+     * HTTP Client for the FlowForge API
```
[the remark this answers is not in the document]
Steve-Mcl: [5334fee](https://github.com/flowforge/flowforge-device-agent/pull/56/commits/5334feec3247e1c1c020f75ed1dc7e90b2ea9880)

lib/AgentManager.js
```diff
@@ -0,0 +1,460 @@
...
+        if (!this.options) {
+            throw new Error('Agent Manager not initialised')
+        }
+        if (this.state === 'starting' || this.state === 'started') {
+            return this.state
+        }
+        this.state = 'starting'
+        info('Agent starting...')
+        try {
+            this.reloadConfig()
```
hardillb, line 70:
```suggestion
            await this.reloadConfig()
```

Steve-Mcl: well spotted fine sir

hardillb (approved):

PR #80, opened 2023-05-03
Ensure log body is a string

hardillb:
part of flowforge/flowforge#2074

## Description

Make sure the message part of any logged content is a string

## Related Issue(s)

flowforge/flowforge#2074

## Checklist

 - [ ] I have read the [contribution guidelines](https://github.com/flowforge/flowforge/blob/main/CONTRIBUTING.md)
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/lib/logging/log.js
+++ b/lib/logging/log.js
@@ -23,6 +23,9 @@ function NRlog (msg) {
         jsMsg = { ts: Date.now(), level: '', msg }
     }
     const date = new Date(jsMsg.ts)
+    if (typeof jsMsg.msg === 'object') {
+        jsMsg.msg = JSON.stringify(jsMsg.msg)
+    }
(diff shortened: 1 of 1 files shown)


lib/logging/log.js
```diff
@@ -23,6 +23,9 @@ function NRlog (msg) {
         jsMsg = { ts: Date.now(), level: '', msg }
     }
     const date = new Date(jsMsg.ts)
+    if (typeof jsMsg.msg === 'object') {
```
knolleary, line 26: Would it be better to invert this? Thinking of other types that don't identify as object that ought to be stringified.

```suggestion
    if (typeof jsMsg.msg !== 'string') {
```

hardillb: Yeah, that makes sense

PR #130, opened 2023-06-30
Ensure agent runs on Node 14

### LINKED ISSUE #123: Add support for Node 14
### Description

Currently the Device Agent requires node 16 or later. That was chosen due to it being the current LTS version at the time of writing and was the right choice.

Whilst Node 14 has reached its end of life, there are a number of devices stuck at that level and likely to be for some time.

We have had a specific request to support Node 14 on the device agent. I haven't sniff tested it yet, but I do not think it will be a large task to get it working. Mostly likely a bit of syntactic sugar will have to be gotten rid of and done the old-fashioned way.

knolleary:
Fixes #123 

## Description

A very minor set of changes needed to remove some Node 16 specific syntax.

Will want to give it a bit more soak testing - but unit tests are clear and initial sniff tests using it all work.

Have also added Node 14 to the unit test runner matrix to ensure we don't regress it given we mostly work in a Node 16+ environment.

## Checklist


knolleary:
/me shakes my fist at eslint and gets to work

Steve-Mcl (approved):
Looks all goodness Nick (bar the failing tests ofc)

Will require an update to FF docs: https://flowforge.com/docs/user/devices/

knolleary:
The underlying cause of the failure, we believe, is the older version of npm in node 14 not installing peer dependencies properly for some of the latest eslint plugins - so this is *only* relevant for running the lint checks.

Fixed by disabling linting on the node 14 run - no technical reason to run it on multiple versions of node.

PR #141, opened 2023-07-12
Adds FF library plugin

Steve-Mcl:
## Description

NOTE: This is part 2 of 2 PRs to enable the feature described in https://github.com/flowforge/flowforge/issues/2294

1. Adds duplicate of `flowforge-nr-launcher/lib/storage/libraryPlugin.js`
2. Checks and stores new boolean property `licensed` to include (or exclude) the necessary settings for loading the plugin
3. Plugin integration is made via `nodesDir` and points to the plugin path within the agent
   1. NOTE: a dummy package.json file was also added alongside the plugin since it would not load without it
4. adds test `updates licensed state`
5. updates existing tests to ensure correct presence and state of the `licensed` flag

## Related Issue(s)

https://github.com/flowforge/flowforge/issues/2294

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


package.json
```diff
@@ -2,6 +2,14 @@
     "name": "@flowforge/flowforge-device-agent",
     "version": "1.9.4",
     "description": "An Edge Agent for running Node-RED instances deployed from the FlowForge Platform",
+    "exports": {
+        "./libraryPlugin": "./lib/plugins/libraryPlugin.js"
+    },
+    "node-red": {
+        "plugins": {
+            "flowforge-library": "lib/plugins/libraryPlugin.js"
```
Steve-Mcl, line 10:
```suggestion
        "./libraryPlugin": "./lib/plugins/node_modules/libraryPlugin.js"
    },
    "node-red": {
        "plugins": {
            "flowforge-library": "lib/plugins/node_modules/libraryPlugin.js"
```

knolleary: I don't think the device agent's package.json should have a node-red section at all - it isn't a module that is intended to provide plugins to anything other than the NR instance it spawns.

Steve-Mcl: Agreed - we can add later if that becomes necessary. Will remove

package.json
```diff
@@ -2,6 +2,14 @@
...
+        }
+    },
```
Steve-Mcl, line 12:
```suggestion
```

Remove "node-red" section - not required right now.

knolleary (approved):
Looks good. Thanks @Steve-Mcl

Issue #145, opened 2023-07-14
Device's Node-RED environment variables: allow controllable propagation of env set to Node-RED process with the hosting operators


elenaviter:
### Description

This is the consequent feature request based on similar in [nr-launcher](https://github.com/flowforge/flowforge-nr-launcher/issues/144).

### Environment variables, sources
In the environment that the Node-RED process operates, there are 2 sources for the environmental variables it will adopt:

- Variables set on the project settings in FlowForge.
- Variables related to FlowForge project identification. These are needed for Node-RED to communicate with the storage endpoints of FlowForge via the storage plugin.

However, there's third which is needed:
- Variables that come into play when the Device instance is deployed via a container orchestrator. These variables can also be introduced in other situations where operators add their environment variables to the instance dynamically (even if this 'orchestrator' is just a docker run on local machine and me who want to set the environment variables for this run).

### Illustration of the need for dynamically provisioned env propagation, to Node-RED
Take for example the scenario where Device instance runs in a Kubernetes environment. By using annotations and labels, you can direct controllers to the appropriate deployment. Kubernetes operators and controllers, in turn, can set custom environments for instances launched under their control. 
A good example of this dynamic provisioning of environmental variables is the linkage of IAM via service account. When a controller of this service account notices a new pod is up and running and linked to a controlled service account, it can mount IAM secrets to the pod and set the environmental variables. Consequently, the software running on the pod is empowered to assume cloud IAM (such as AWS, Azure, etc.) and operate seamlessly, transparently, via linking the web identity by controllers in form of env variables.

Please note that this is only one example of how environmental variables can be **dynamically provisioned**. There are variables and values which cannot be set from the Snapshot.

To forward these environment variables to Node-RED (as well as software solutions running in Node-RED), we need to imbue the launcher with the ability to augment the application's environment with `process.env`. 

Instead of the usual [way that Node-RED process will assume env](https://github.com/elenaviter/flowforge-device-agent/blob/917231bf7c0e458b847201069e8aa278482c8f4c/lib/launcher.js#L263-L266) includes:
- env set on Snapshot
- env with FF related project identification needed for Node-RED to communicate to storage via storage plugin

E.g. instead
```javascript
const env = Object.assign({}, this.snapshot.env)
if (this.settings?.env) {
    Object.assign(env, this.settings?.env)
}
```

We propose to filter the `process.env` to exclude variables starting with "FORGE", which stands for servicing FORGE secrets:
```javascript
[...]

PR #146, opened 2023-07-14
Device Node-RED app environment variables: allow controllable propagation of env set to process to Node-RED app

elenaviter:
## Description

This is the consequent feature request based on similar in [nr-launcher](https://github.com/flowforge/flowforge-nr-launcher/issues/144).

### Environment variables, sources
In the environment that the Node-RED process operates, there are following sources for the environmental variables:

1. Variables set on the snapshot.
3. Variables that come into play when the Device instance is deployed via a container orchestrator (dynamically provisioned).

### Share env variables to software running with Node-RED
To forward environment variables to Node-RED (as well as software solutions running in Node-RED), we need to imbue the launcher with the ability to augment the application's environment with env from source 3 - e.g. those that are set on `process.env`. 

### Way to share env
Instead of the usual [way that Node-RED process will assume env](https://github.com/elenaviter/flowforge-device-agent/blob/917231bf7c0e458b847201069e8aa278482c8f4c/lib/launcher.js#L263-L266) includes:
- env set on Snapshot
- env with FF related project identification needed for Node-RED to communicate to storage via storage plugin

E.g. instead
```javascript
const env = Object.assign({}, this.snapshot.env)
if (this.settings?.env) {
    Object.assign(env, this.settings?.env)
}
```

We propose to propagate also `process.env`, while filter it to exclude variables starting with "FORGE", which stands for servicing FORGE secrets:
```javascript
const filter_env = (env) =>
    Object.entries(env).reduce((acc, [key, value]) =>
        key.startsWith('FORGE') ? acc : {...acc, [key]: value}, {});
const env = Object.assign({}, this.snapshot.env, filter_env(process.env))
if (this.settings?.env) {
    Object.assign(env, this.settings?.env)
}
```

## Related Issue(s)

[Device's Node-RED environment variables: allow controllable propagation of env set to Node-RED process with the hosting operators](https://github.com/flowforge/flowforge-device-agent/issues/145)

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/lib/launcher.js
+++ b/lib/launcher.js
@@ -260,7 +260,10 @@ class Launcher {
             await this.writeThemeFiles()
         }
 
-        const env = Object.assign({}, this.snapshot.env)
+        const filter_env = (env) =>
+            Object.entries(env).reduce((acc, [key, value]) =>
+                key.startsWith('FORGE') ? acc : {...acc, [key]: value}, {});
+        const env = Object.assign({}, this.snapshot.env, filter_env(process.env))
         if (this.settings?.env) {
             Object.assign(env, this.settings?.env)
         }
--- a/lib/launcher.js
+++ b/lib/launcher.js
@@ -260,10 +260,15 @@ class Launcher {
             await this.writeThemeFiles()
         }
 
-        const filter_env = (env) =>
+        const filterEnv = (env) =>
             Object.entries(env).reduce((acc, [key, value]) =>
-                key.startsWith('FORGE') ? acc : {...acc, [key]: value}, {});
-        const env = Object.assign({}, this.snapshot.env, filter_env(process.env))
+                key.startsWith('FORGE') ? acc : { ...acc, [key]: value }, {})
+
+        // According to https://github.com/flowforge/flowforge-nr-launcher/pull/145,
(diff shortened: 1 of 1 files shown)


knolleary:
I think this feature needs to have a flag to enable it on the device. We don't want, by default, to pass through the env vars.

The device is a difference situation to the launcher. The end user has full access to the device and can modify it to do anything they want.

For this, I propose this gets enabled by setting a flag in the `device.yml` configuration file - something that has to be done as part of setting up the device.

Would that fit your needs?

knolleary (approved):
Good first iteration. Depending on feedback we may also want a way to enable this via the `device.yml` file to simplify the configuration, but we can run with this for now.

elenaviter:
@knolleary, greetings once again!
Regarding your suggested approach for a more controlled configuration:

> I recommend enabling this via a flag in the device.yml configuration file, which would be part of the device setup.

This makes a lot of sense to me for the next phase. 
We can create a separate issue to address this. What would you say?

knolleary:
> We can create a separate issue to address this. What would you say?

Yes please!

PR #152, opened 2023-07-25
Move plugin location so it can be loaded by NR 3.0.x

knolleary:
## Description

Having tested #141 locally against NR 3.0.2, I couldn't get the library plugin to load. Some changes have gone into NR 3.1 in this area, but as that is not generally available yet, we need a solution that is 3.0.x compatible.

The plugin loading logic searches the the tree expecting it to be in a `node_modules` directory, otherwise it ignores it. This PR moves the plugin files under this structure so it appears in the right place.

I have tested this against both 3.0.2 and 3.1.0-beta.3


Steve-Mcl (approved):
makes sense.

PR #157, opened 2023-08-09
Fix device running or reporting old snapshot id/name

Steve-Mcl:
## Description

This PR addresses several issues found while debugging the reported problem.

NOTE: None of the work here involved messing with the device files. The state of the device and its files were all a result of actions taken in the forge application.  I feel this is important to point out because as it stands, it is possible to get the device to crash and into odd states.

The issues encountered...

### problem 1 - device running snapshot of deleted instance
Device active/target snapshot get populated in checkin code
device reports it is on old snapshot & that matches 
checkin code did not check that the activeSnapshots project matched the device current project

### problem 2 - snapshot table is not cleaned - holds snapshots from deleted instances
When device checks in, its reported snapshot is found and the device objects active/target snapshot are populated meaning the device passes its checkin & does not instruct an update.

NOTE: No work was done to delete the old snapshots, only to mitigate the problem. A separate issue will be raised detailing thoughts and care points around this.  

### problem 3
Even if we force a DeviceUpdate, it just passes the exact same settings due to the snapshot table having old entries

### problem 4
Once we do get the device to clean up, switching between dev/autonomous mode caused a crash due to missing files
```
[AGENT] 09/08/2023 09:47:27 [info] Cleaning instance directory
[AGENT] 09/08/2023 09:47:50 [info] Enabling developer mode
[AGENT] 09/08/2023 09:47:51 [info] Enabling remote editor access
[AGENT] 09/08/2023 09:47:51 [info] No running Node-RED instance, not starting editor
[AGENT] 09/08/2023 09:47:56 [info] Enabling remote editor access
[AGENT] 09/08/2023 09:47:56 [info] No running Node-RED instance, not starting editor
[Error: ENOENT: no such file or directory, open 'C:\opt\flowforge-device\project\flows.json'] {
  errno: -4058,
  code: 'ENOENT',
  syscall: 'open',
  path: 'C:\\opt\\flowforge-device\\project\\flows.json'
}
```

### problem 5
when the agent determines a new snapshot is available, we dont instruct it to grab new settings.
Since some env vars are computed by the platform (e.g. FF_SNAPSHOT_ID/NAME) we must also instruct
the agent to refresh its settings.

#### 1: [Improve detection of changes after dev mode](https://github.com/flowforge/flowforge-device-agent/commit/385e5ede2844d21364fad004e3f6a60135fc5b7c)

Here, instead of doing a flow comparison (potentially costly operation & potentially pointless if the new snapshot flows match the old snapshot flows), the checks are broken down into 3 individual parts (each increasing in processing requirements)
* check snapshot IDs for change
* check system env vars for change (e.g. has FF_SNAPSHOT_ID changed?)
* check flows for change

If any have changed, we set a flag to indicate a reload is necessary

#### 2: [skip to updates when change detected](https://github.com/flowforge/flowforge-device-agent/commit/edce449b274e1787770b84c4c0345c542d28d048)

Under certain circumstances, it is possible receive an update WITHOUT a snapshot. In this case, the `setState` was entering the middle of the state machine and NOT getting to the important part where updates are applied.

#### 3: [ensure settings update if snapshot needs update](https://github.com/flowforge/flowforge-device-agent/commit/29f9ec01fb48f322c1deac218272fea3ba70b3c4)

Since FF_SNAPSHOT_ID was added we _should_ **always** reload the settings when the snapshot changes.
Failing to do so means we hold stale info in the env vars.

#### 4: [improve feedback info: include snapshot id/name](https://github.com/flowforge/flowforge-device-agent/commit/5313a584d743ac605e6c7fa6d5652b5f0991681e)

Specifically 
- deduplicate the 2 places where Instance, Snapshot, Project etc are logged to console into a common routine. 
- adds the FF_* env vars to the log output (so user can SEE when the env vars are out of step with the config)

#### 5: [update tests](https://github.com/flowforge/flowforge-device-agent/commit/5313a584d743ac605e6c7fa6d5652b5f0991681e)

2 new tests added
- reloads latest snapshot from platform when switching off developer mode (if snapshot ID changed)
- reloads latest snapshot from platform when switching off developer mode (if platform env vars are modified)

Additionally: Fixes several failing tests due to now ALWAYS reloading settings if the snapshot has changed (see item 3 above)

## Related Issue(s)

#132 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -53,14 +53,27 @@ class Agent {
                     this.currentMode = config.mode || 'autonomous'
                     this.config.licensed = config.licensed || null
                 }
-                info(`Instance      : ${this.currentProject || 'unknown'}`)
-                info(`Snapshot      : ${this.currentSnapshot?.id || 'none'}`)
-                info(`Settings      : ${this.currentSettings?.hash || 'none'}`)
-                info(`Operation Mode: ${this.currentMode || 'unknown'}`)
-                info(`Licensed      : ${this.config.licensed === null ? 'unknown' : this.config.licensed ? 'yes' : 'no'}`)
+                this.printAgentStatus()
             } catch (err) {
                 warn(`Invalid project file: ${this.projectFilePath}`)
             }
+        }
+    }
+
+    printAgentStatus () {
+        info('Configuration :-')
+        info(`  * Instance         : ${this.currentProject || 'unknown'}`)
+        info(`  * Snapshot         : ${this.currentSnapshot?.id || 'none'}`)
+        info(`  * Settings         : ${this.currentSettings?.hash || 'none'}`)
+        info(`  * Operation Mode   : ${this.currentMode || 'unknown'}`)
+        info(`  * Licensed         : ${this.config.licensed === null ? 'unknown' : this.config.licensed ? 'yes' : 'no'}`)
+        if (typeof this.currentSettings?.env === 'object') {
+            info('Environment :-')
+            info(`  * FF_SNAPSHOT_ID   : ${this.currentSettings.env.FF_SNAPSHOT_ID || ''}`)
+            info(`  * FF_SNAPSHOT_NAME : ${this.currentSettings.env.FF_SNAPSHOT_NAME || ''}`)
+            info(`  * FF_DEVICE_ID     : ${this.currentSettings.env.FF_DEVICE_ID || ''}`)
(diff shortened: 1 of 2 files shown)


hardillb:
Stopped the device, deleted the instance, started the device. Device comes up in stopped state which feels right

hardillb:
Stopped again, re-assigned to new instance, started, loads new instance snapshot

Issue #162, opened 2023-08-21
Docker Device Agent failed to pull snapshot after restart


robmarcer:
### Current Behavior

After a nightly restart of the Device Agent running on Docker, the process of getting the snapshot timed out.

![Annotation-Annotation on 2023-08-21 at 16-20-14 png](https://github.com/flowforge/flowforge-device-agent/assets/92092324/e79fd33e-c445-4377-90a6-de66aa2726ce)

### Expected Behavior

The Device agent should keep retrying the snapshot download if the prior attempt timed out.

### Steps To Reproduce

The network this device is running on is believed to have restarted at the same time as the device agent.

### Environment

- FlowForge version: FF Cloud 1.10.2
- Node.js version: 16.20.2
- npm version:
- Platform/OS:
- Browser:

robmarcer:
I've done some testing on this issue. I've used a slow network simulator (https://www.sitespeed.io/documentation/throttle/)to test what happens if the device agent can't get the snapshot within the 10000ms limit.

It seems that if it fails to get the snapshot in time the following error is shown:

[AGENT] 8/24/2023 8:24:59 PM [warn] Problem getting snapshot: TimeoutError: Timeout awaiting 'request' for 10000ms

It seems that the device agent then just hangs.

In my testing I created a 3MB snapshot (it contains 3 inject nodes which each have a 1MB string in them) and slowed my network down to 50kb up, 50kb down.

Video shows the test - https://d.pr/v/vGRjJR

hardillb:
OK, but 50kps is very slow by modern network standards.

What timeout should be use? We could double the current 10 second would that be enough?

robmarcer:
Thanks @hardillb, this is a great fix for now. As discussed, I think there is something off with the retry logic for this particular timeout. It seems that once the snapshot download fails once the system doesn't retry, where as various other scenarios I tested seem to periodically repeat the failed action.

hardillb:
How big are their snapshots, and just how slow is their network?

robmarcer:
Snapshots are 280KB (well slighting more as that number is for the flow not the full snapshot), ~20 device agents running on the same machine so ~5.6MB of data to pull.

Anything under a ~1.5Mb connection would result in a timeout.

hardillb:
So they are effectively DoSing themselves by mass restarting them all at the same time.

The "right" solution may be that they stagger the restart....

knolleary:
[...]

PR #163, opened 2023-08-25
Increase snapshot download timeout

hardillb:
fix for #162

## Description

Slow network is causing larger snapshots to timeout when being downloaded.

Doubled the download time just for snapshots, might still want to look at retry on failure.

## Related Issue(s)

#162

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/lib/http.js
+++ b/lib/http.js
@@ -31,7 +31,7 @@ class HTTPClient {
 
     async getSnapshot () {
         try {
-            return await this.client.get('live/snapshot').json()
+            return await this.client.get('live/snapshot', {timeout: {request: 20000 }}).json()
         } catch (err) {
             warn(`Problem getting snapshot: ${err.toString()}`)
             debug(err)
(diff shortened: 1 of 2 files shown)


lib/http.js
```diff
@@ -31,7 +31,7 @@ class HTTPClient {
 
     async getSnapshot () {
         try {
-            return await this.client.get('live/snapshot').json()
+            return await this.client.get('live/snapshot', { timeout: { request: 20000 } }).json()
```
Steve-Mcl, line 34: In the absence of any user settable property (where they can manage the timeout value themselves), is there any reason we dont crank this a little further? like 30s or 45s ?

hardillb: I didn't want to encourage even bigger flows... but we could go to 30 i suppose.

Steve-Mcl:
```suggestion
            return await this.client.get('live/snapshot', { timeout: { request: 30000 } }).json()
```

I like 30 :)

Steve-Mcl (approved):

PR #169, opened 2023-09-07
Do not attempt editor reconnect if platform tells us No Tunnel

### LINKED ISSUE #168: Editor tunnel gets into retry-loop if initial connection timesout
### Current Behavior

When the device agent is asked to create its websocket tunnel connection back to the platform, it has 5 seconds to complete and send-back an acknowledgement.

If the connection happens after the timeout occurs, the platform closes the connection as it doesn't think it should exist. The device agent doesn't know this - so starts retrying to connect.

However at this point, the device agent internal state has got out of sync and the code to handle breaking out of the retry loop doesn't kick in.

### Expected Behavior

Better handling if the initial connect timesout. Not 100% what the ultimately right behaviour is... just that the current behaviour isn't right.

### Steps To Reproduce

This was originally reported by a user of FF Cloud - where their device is on a slow network connection.

To reproduce locally I added a `setTimeout()` to introduce an artificial delay in handling the `startEditor` command message.

### Environment

- FlowForge version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

knolleary:
Fixes #168 

## Description

If the connection is too slow to create, the platform will close the connection and discard the session token. The device agent just sees a connection drop so automatically retries.

But as the token has been discarded, it isn't going to succeed - but keeps trying anyway.

This fix checks the reason code the platform provided (if any). The 'No tunnel' reason tells us the platform doesn't have a matching session token so there is no point retrying.

### DIFF
--- a/lib/editor/tunnel.js
+++ b/lib/editor/tunnel.js
@@ -202,21 +202,29 @@ class EditorTunnel {
             })
         }
         socket.on('close', async (code, reason) => {
+            if (Buffer.isBuffer(reason)) {
+                reason = reason.toString()
+            }
             // The socket connection to the platform has closed
             info(`Editor tunnel closed code=${code} reason=${reason}`)
             socket.removeAllListeners()
             this.socket = null
             clearTimeout(this.reconnectTimeout)
-
-            // Assume we need to be reconnecting. If this close is due
-            // to a request from the platform to turn off editor access,
-            // .close will get called
-            const reconnectDelay = this.reconnectDelay
-            // Bump the delay for next time... 500ms, 1.5s, 4.5s, 10s 10s 10s...
-            this.reconnectDelay = Math.min(this.reconnectDelay * 3, 10000)
-            this.reconnectTimeout = setTimeout(() => {
-                this.connect()
-            }, reconnectDelay)
+            if (reason !== 'No tunnel') {
+                // Assume we need to be reconnecting. If this close is due
+                // to a request from the platform to turn off editor access,
+                // .close will get called
+                const reconnectDelay = this.reconnectDelay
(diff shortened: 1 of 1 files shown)


lib/editor/tunnel.js
```diff
@@ -202,21 +202,29 @@ class EditorTunnel {
...
-            // Assume we need to be reconnecting. If this close is due
-            // to a request from the platform to turn off editor access,
-            // .close will get called
-            const reconnectDelay = this.reconnectDelay
-            // Bump the delay for next time... 500ms, 1.5s, 4.5s, 10s 10s 10s...
-            this.reconnectDelay = Math.min(this.reconnectDelay * 3, 10000)
-            this.reconnectTimeout = setTimeout(() => {
-                this.connect()
-            }, reconnectDelay)
+            if (reason !== 'No tunnel') {
```
Steve-Mcl, line 213: What generates this text Nick?

Just thinking, if it is our code in the platform (or even if it is generated in socket package), a change of text would bypass this check.

Should/could we use a well defined code instead of reason text?

knolleary: This text does come from the platform and is set by us. I agree it's a little brittle. Will consider a change of code - need to look at what the ws spec allows.

knolleary: Have updated the PR to look for either:

 - `1008/No tunnel` - for FF 1.11 and earlier
 - `4004` - for FF 1.12

We are using `100*` codes in multiple places where we should not be - they are all protocol specific reserved codes we shouldn't be overloading. I have made some minor changes in the platform PR to use 400* range codes which are the ones reserved for application specific errors. Will need to do a broader tidy up.

Steve-Mcl (approved):

PR #171, opened 2023-09-12
Create raspbian install script

MarianRaphael:
## Description

Creation of a Raspbian install script

## Checklist

 - [ ] I have read the [contribution guidelines](https://github.com/flowforge/flowforge/blob/main/CONTRIBUTING.md)
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/service/raspbian-install-device-agent.sh
+++ b/service/raspbian-install-device-agent.sh
@@ -1,0 +1,109 @@
+#!/bin/bash
+
+if [[ $EUID -ne 0 ]]; then
+    echo "This script must be run as root"
+    exit 1
+fi
+
+# Node.js version
+MIN_NODEJS=16
+
+# Update package list and upgrade installed packages
+apt-get update
+
+# Helper functions to test for existence of npm
+function HAS_NPM {
+    if [ -x "$(command -v npm)" ]; then return 0; else return 1; fi
+}
+# Install Node.js and npm if they do not exist
+if [ -x "$(command -v node)" ]; then
+    echo "NodeJS found"
+    VERSION=$(node --version | cut -d "." -f1 | cut -d "v" -f2)
+
+    if [[ $VERSION -ge $MIN_NODEJS ]]; then
+        echo "**************************************************************"
+        echo " NodeJS Version $MIN_NODEJS or newer found "
+        echo "**************************************************************"
+    else
+        echo "**************************************************************"
+        echo " You need NodeJS $MIN_NODEJS or newer, please upgrade "
+        echo "**************************************************************"
+        exit 1
+    fi
+else
+    echo "**************************************************************"
+    echo " No NodeJS found"
+    echo " Do you want to install NodeJS $MIN_NODEJS?"
+    echo "**************************************************************"
+    read -p "y/N: " yn
+    [ -z "$yn" ] && yn="n"
+
+    if [[ "$yn" == "y" ]] || [[ "$yn" == "Y" ]]; then
+        if [ ! -x "$(command -v curl)" ]; then
+            sudo apt-get install -y curl
+        fi
+        sudo apt-get install -y ca-certificates curl gnupg
+        sudo mkdir -p /etc/apt/keyrings
+        curl -fsSL https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | sudo gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg
+        echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesource.com/node_$MIN_NODEJS.x nodistro main" | sudo tee /etc/apt/sources.list.d/nodesource.list
+        sudo apt-get update
+        sudo apt-get install nodejs -y
+    else
+        echo "**************************************************************"
+        echo " You will need to manually install NodeJS first. "
+        echo " Exiting."
+        echo "**************************************************************"
+        exit 1
+    fi
+fi
+
+if ! HAS_NPM; then
+    # User confirmation for installation
+    read -p "Do you wish to proceed with the installation of npm? (y/n): " proceed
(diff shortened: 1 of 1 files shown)


MarianRaphael:
Is script the right folder to place an installation helper script in?
Do we want such a script in the repo? Otherwise, I would just include it in docs?

service/raspbian-install-device-agent.sh
```diff
@@ -0,0 +1,109 @@
+#!/bin/bash
+
+if [[ $EUID -ne 0 ]]; then
+    echo "This script must be run as root"
```
hardillb, line 4: The flowfuse install script uses sudo for any actions that need root. On raspbian this works without needing a password, but the script will ask for password if needed

https://github.com/flowforge/installer/blob/a7909d1074106fa3f12abf3c8a4241eb6dcda44b/install.sh#L6

service/raspbian-install-device-agent.sh
```diff
@@ -0,0 +1,109 @@
...
+    else
+        echo "**************************************************************"
+        echo " You need NodeJS $MIN_NODEJS or newer, please upgrade "
+        echo "**************************************************************"
+        exit 1
+    fi
+else
+    echo "**************************************************************"
+    echo " No NodeJS found"
+    echo " Do you want to install NodeJS $MIN_NODEJS?"
```
hardillb, line 36: This should probably install at least NodeJS 18 if installing from new

service/raspbian-install-device-agent.sh
```diff
@@ -0,0 +1,109 @@
+#!/bin/bash
+
+if [[ $EUID -ne 0 ]]; then
+    echo "This script must be run as root"
+    exit 1
+fi
+
+# Node.js version
+MIN_NODEJS=16
```
hardillb, line 9: We technically support NodeJS 14 for the device agent

service/raspbian-install-device-agent.sh
```diff
@@ -0,0 +1,109 @@
...
+        sudo apt-get update
+        sudo apt-get install nodejs -y
+    else
+        echo "**************************************************************"
+        echo " You will need to manually install NodeJS first. "
+        echo " Exiting."
+        echo "**************************************************************"
+        exit 1
+    fi
+fi
```
hardillb, line 58: This won't work on ARMv6 builds as the standard NodeJS builds don't support it. We should document that Pi Zero's are not supported

service/raspbian-install-device-agent.sh
```diff
@@ -0,0 +1,109 @@
...
+[Install]
+WantedBy=multi-user.target" | sudo tee /etc/systemd/system/flowforge-device-agent.service >/dev/null
+
+# Reload systemd, enable and start the service
+sudo systemctl daemon-reload
+sudo systemctl enable flowforge-device-agent.service
+sudo systemctl start flowforge-device-agent.service
+
+# Output status of the service
+sudo systemctl status flowforge-device-agent.service
```
hardillb, line 109: We should either run all commands with sudo or none (assume the whole script is running as root)

hardillb (commented):
Looks good, just a few comments

service/raspbian-install-device-agent.sh
MarianRaphael:
```suggestion
MIN_NODEJS=14
```

service/raspbian-install-device-agent.sh
MarianRaphael:
```suggestion
    echo " Do you want to install NodeJS 18?"
```

service/raspbian-install-device-agent.sh
```diff
@@ -0,0 +1,109 @@
...
+    [ -z "$yn" ] && yn="n"
+
+    if [[ "$yn" == "y" ]] || [[ "$yn" == "Y" ]]; then
+        if [ ! -x "$(command -v curl)" ]; then
+            sudo apt-get install -y curl
+        fi
+        sudo apt-get install -y ca-certificates curl gnupg
+        sudo mkdir -p /etc/apt/keyrings
+        curl -fsSL https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | sudo gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg
+        echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesource.com/node_$MIN_NODEJS.x nodistro main" | sudo tee /etc/apt/sources.list.d/nodesource.list
```
MarianRaphael, line 48:
```suggestion
        echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesource.com/node_18.x nodistro main" | sudo tee /etc/apt/sources.list.d/nodesource.list
```

service/raspbian-install-device-agent.sh
MarianRaphael: I will add it to the doc: https://flowfuse.com/docs/hardware/introduction/

service/raspbian-install-device-agent.sh
MarianRaphael: see [added sudo](https://github.com/FlowFuse/device-agent/pull/171/commits/0fb2eac669107100c6211ff69a0a3f0fb308a1fb)

service/raspbian-install-device-agent.sh
hardillb: Now all the commands are being run with `sudo` we shouldn't need  this warning, but we should swap the test to the negative and show a message saying it may ask for a password to run sudo only if not being run as root.

hardillb:
@MarianRaphael Looks good, just needs the warning at the start updating to say it may ask for user password for sudo if not run as root.

MarianRaphael:
@hardillb I updated the script and added the info text

hardillb (approved):

Issue #178, opened 2023-09-27
Node-RED is not updated if package lock exists


Steve-Mcl:
### Current Behavior

When switching snapshots with different versions of Node-RED specified, "Installing Packages" is ran but Node-RED version remains old.
`npm install --production` (as the launcher and agent execute) ignore `package.json` and respect the lock file instead

### Expected Behavior

We should either we be inspecting the package or indiscriminately deleting the lock file in the agent device project dir when "Updating configuration files"-> writePackage() is called

### Steps To Reproduce

Only repeatable with new Deploy Snapshot to App Owned Devices.

### Environment

- FlowForge version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

Pezmc:
`npm install --production` should always read the package.json and update the package-lock. So I think there must be something else at the root of why the NR package for example isn't updating.

It's `npm ci` that is intended to read the package lock and fail the command if the package-lock and package.json aren't in sync.

Testing locally:

```json5
// package.json
{
  "name": "testing",
  "dependencies": {
    "node-red": "^3.0.0"
  }
}
```

```shell
$ npm install
$ ls
node_modules/      package-lock.json  package.json
$ cat node_modules/node-red/package.json | jq .version
"3.0.0"
```


```json5
// Updated package.json
{
  "name": "testing",
  "dependencies": {
    "node-red": "^3.1.0"
  }
}
```

```shell
$ npm install --production
npm WARN config production Use `--omit=dev` instead.

up to date, audited 299 packages in 617ms

$ cat node_modules/node-red/package.json | jq .version
"3.1.0"
```

Steve-Mcl:
@Pezmc thanks for taking the time to reason this out. 

I can only say, from what I witnessed what was occurring. And deleting package lock fixed the issue.

Perhaps it was an NPM version thing, or a windows thing, or a `~^>=` thing, or something else entirely.  The only thing I can say without re-testing and capturing more details is that deleting the lock, this stopped happening 100%.

---

_Brain engaged_

This might have been when changing stack from NR 3.1 to NR3.0.2.  But I was certain it was in both upgrade and downgrade situations.

Pezmc:
If it was only for the _downgrade_, that could explain things, 3.1.0 is semantically compatible with 3.0.2, so it wouldn't downgrade the package.

[...]

PR #179, opened 2023-09-27
delete lock file when package is updated

Steve-Mcl:
## Description

Deletes package lock when the package is updated.
This permits the internal call to `npm install --production` to get specified versions instead of the lock versions.

### TODO:

- [x] test package is deleted.

## Related Issue(s)

#178 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/lib/launcher.js
+++ b/lib/launcher.js
@@ -33,6 +33,7 @@ class Launcher {
 
         this.files = {
             packageJSON: path.join(this.projectDir, 'package.json'),
+            packageLockJSON: path.join(this.projectDir, 'package-lock.json'),
             flows: path.join(this.projectDir, 'flows.json'),
             credentials: path.join(this.projectDir, 'flows_cred.json'),
             settings: path.join(this.projectDir, 'settings.js'),
@@ -54,6 +55,13 @@ class Launcher {
         await fs.writeFile(this.files.packageJSON, JSON.stringify(packageData, ' ', 2))
         await fs.rm(path.join(this.projectDir, '.config.nodes.json'), { force: true })
         await fs.rm(path.join(this.projectDir, '.config.nodes.json.backup'), { force: true })
+
+        // as the package.json file has been re-written, we will "assume" different versions
+        // of the modules and therefore remove the package-lock.json file to permit modules
+        // to be updated to the versions specified in package.json
+        if (existsSync(this.files.packageLockJSON)) {
+            await fs.rm(this.files.packageLockJSON)
+        }
     }
 
     async readPackage () {
--- a/lib/launcher.js
+++ b/lib/launcher.js
@@ -59,6 +59,10 @@ class Launcher {
         // as the package.json file has been re-written, we will "assume" different versions
         // of the modules and therefore remove the package-lock.json file to permit modules
         // to be updated to the versions specified in package.json
+        await this.removePackageLock()
+    }
+
+    async removePackageLock () {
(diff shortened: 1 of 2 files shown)


lib/launcher.js
```diff
@@ -54,6 +55,13 @@ class Launcher {
         await fs.writeFile(this.files.packageJSON, JSON.stringify(packageData, ' ', 2))
         await fs.rm(path.join(this.projectDir, '.config.nodes.json'), { force: true })
         await fs.rm(path.join(this.projectDir, '.config.nodes.json.backup'), { force: true })
+
+        // as the package.json file has been re-written, we will "assume" different versions
+        // of the modules and therefore remove the package-lock.json file to permit modules
+        // to be updated to the versions specified in package.json
+        if (existsSync(this.files.packageLockJSON)) {
+            await fs.rm(this.files.packageLockJSON)
```
Pezmc, line 63: Wouldn't an `npm install` already handle this? 

Or are you wanting to run `npm upgrade` to upgrade the package log speficially?

Steve-Mcl: Pez, was discussed in slack

> 
> Steve
> When switching snapshots with different versions of Node-RED specified, "Installing Packages" is ran but Node-RED version remains old.
> npm install --production (as the launcher and agent execute) ignore package.json and respect the lock file instead
> Should we be inspecting the package or indiscriminately deleting the lock file in the agent device project dir when "Updating configuration files"-> writePackage() is called?
> 2 replies
> 
> 
> nick
> My instinct is to delete the package lock file whenever we update the package file
> 
> 
> Steve

> same - but wanted to check

Pezmc: To me, this could lead to unexpected upgrades outside of the package changes, but so would running an explicit npm upgrade, so unblocking until we find a better solution for now.

lib/launcher.js
```diff
@@ -54,6 +55,17 @@ class Launcher {
         await fs.writeFile(this.files.packageJSON, JSON.stringify(packageData, ' ', 2))
         await fs.rm(path.join(this.projectDir, '.config.nodes.json'), { force: true })
         await fs.rm(path.join(this.projectDir, '.config.nodes.json.backup'), { force: true })
+
+        // as the package.json file has been re-written, we will "assume" different versions
+        // of the modules and therefore remove the package-lock.json file to permit modules
```
Pezmc, line 60: It may be worth noting that this is only because of the `--production` flag, a normal `npm install` would update the package-lock.

Pezmc (approved):

lib/launcher.js
Pezmc: On second thought, this still doesn't fully grok for me actually, an `npm install` always updates the package-lock based on the latest package.json as part of the install. Testing locally.

Pezmc: Result of my tests in https://github.com/FlowFuse/device-agent/issues/178#issuecomment-1750306065

Pezmc (changes requested):
I don't think this gets to the bottom of why the packages aren't updating, see comment on the original issue: https://github.com/FlowFuse/device-agent/issues/178#issuecomment-1750306065

Steve-Mcl:
@Pezmc @knolleary 

Good news: I cannot recreate this from staging to device.  On windows or Linux.  

I can only assume some other mis-op (by me)

I will park (close) this PR (but not delete the branch for now)

PR #189, opened 2023-10-25
Ensure that tunneled payloads are parsed

hardillb:
## Description

HTTP Body is JSON.stringify() before being sent down the tunnel need to JSON.parse() before forwarding on to editor

## Related Issue(s)

flowfuse/flowfuse#2999

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/lib/editor/tunnel.js
+++ b/lib/editor/tunnel.js
@@ -169,10 +169,11 @@ class EditorTunnel {
                     // ↓ useful for debugging but very noisy
                     // console.log('Making a request to:', fullUrl, 'x-access-token:', request.method, reqHeaders['x-access-token'])
                     // debug(`proxy [${request.method}] ${fullUrl}`)
+                    // debug(`body ${request.body}`)
                     const options = {
                         headers: reqHeaders,
                         method: request.method,
-                        body: request.body,
+                        body: request.body ? JSON.parse(request.body) : undefined,
                         throwErrors: false
                     }
                     if (thisTunnel.localProtocol === 'https') {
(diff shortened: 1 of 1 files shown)


hardillb:
For reference here is the other end of the tunnel `JSON.stringify()` call

knolleary:
With this change, I can load the device editor, but I cannot deploy any changes

hardillb:
When you get a second can you look at this revised fix, shouldn't break anything this time and still allow UIBuilder to work.

hardillb:
@Steve-Mcl If you get a moment can you kick the tires on this please. I've given it a better test than the first version and I'm 99% sure it won't break anything this time.

lib/editor/tunnel.js
```diff
@@ -169,10 +169,20 @@ class EditorTunnel {
...
+                    // debug(`body ${request.body}`)
+                    let body = request.body
+                    try {
+                        const temp = JSON.parse(body)
+                        if (typeof temp === 'string') {
+                            body = temp
+                        }
+                    } catch (err) {
+                        // ignore
+                    }
```
knolleary, line 181:
```suggestion
                    // debug(`body ${request.body}`)
                    let body = request.body
                    if (body && body[0] === '"') {
                        try {
                            body = JSON.parse(body)
                        } catch (err) {
                            // ignore
                        }
                    }
```

This feels an expensive way to only parse strings. Suggest adding a check to see if it looks like a JSON encoded String before attempting the parse.

knolleary (changes requested):

knolleary (approved):

PR #197, opened 2023-11-08
Move npm org

knolleary:
## Description

This updates the Device Agent to the new `@flowfuse/device-agent` npm package and `flowfuse/device-agent` dockerhub image.

As part of this migration it also:

 - Updates the readme to cover the new locations
 - Updates the default data directory to `/opt/flowfuse-device/` - however it will use `/top/flowforge-device` if it exists and nothing else has been specified
 - Updates the workflow to cross-publish to the existing npm package and container images. This required a dedicated Dockerfile to ensure the proper config directory still exists for the 'legacy' container.
 - Adds a startup log message if found to be installed using the old npm package name

The full set of workflow changes are untested - we cannot test it without doing a full release. So please give the workflow changes a particularly close review.

Before we do the first release once merged, we need to update some repo-level secrets:

 - `NPM_PUBLISH_TOKEN_FLOWFORGE` - the token for publishing to old npm org
 - `NPM_PUBLISH_TOKEN` - the token for publishing to the new org
 - `DOCKER_HUB_PASSWORD_FLOWFORGE` - the password for the old dockerhub account
 - `DOCKER_HUB_PASSWORD` - the password for the new dockerhub account

### DIFF
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -60,7 +60,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions/setup-node@v1
+      - uses: actions/setup-node@v3
         with:
           node-version: 16
       - name: "Install jq"
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -57,6 +57,7 @@ jobs:
           password: ${{ secrets.DOCKER_HUB_PASSWORD }}
           readme-filepath: ./docker/README.md
   publish_legacy:
+    needs: [publish]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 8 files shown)


knolleary:
@hardillb asking you for a review here - in particular, the workflow steps to cross publish.

hardillb:
Looking

hardillb (approved):
My only worry is that the old package may not be fully processed by npmjs.org before we build the container.

We had the wait action at the start that was commented out, it may need to go in the legacy path.

.github/workflows/release-publish.yml
```diff
@@ -47,20 +39,72 @@ jobs:
...
+      - name: Publish README.md to Docker hub
+        uses: peter-evans/dockerhub-description@v3
+        with:
+          repository: flowfuse/device-agent
+          username: flowfuse
+          password: ${{ secrets.DOCKER_HUB_PASSWORD }}
+          readme-filepath: ./docker/README.md
+  publish_legacy:
+    runs-on: ubuntu-latest
+    steps:
```
knolleary, line 61:
```suggestion
  publish_legacy:
    needs: [publish]
    runs-on: ubuntu-latest
    steps:
```

knolleary: @hardillb this should ensure the legacy build only runs once the publish of the new package has happened; the legacy container uses the new package internally.

hardillb: Ahh, yes because the legacy container is still using the new package. makes sense.

PR #205, opened 2023-11-14
Include project nodes for app assigned devices

Steve-Mcl:
## Description

This pull request adds project nodes for app assigned devices. The `@flowfuse/nr-project-nodes` package has been added to the `package.json` file, and the `Launcher` class has been updated to include the path to project nodes. NOTE, the `settings.nodesDir` (used to provide access to the project nodes for Node-RED) is conditionally added based on the env being EE and will include a different path to the `@flowfuse/nr-project-nodes` package in the dev-env when detected.

This pull request addresses the following commits:

- [Include project nodes for app assigned device](https://github.com/FlowFuse/device-agent/commit/fa02e707e795995db0f68e67badfe8312a7f61f1)
- [Fix nodesDir path in launcher.js](https://github.com/FlowFuse/device-agent/commit/86d1ada1a82c51c50bb5516a43b517cb999cdde2)
- [Refactor package path handling to support testing](https://github.com/FlowFuse/device-agent/commit/4d2f7d639f599869b468b5729cb1e151707aa7e5)

## Related Issue(s)

https://github.com/FlowFuse/flowfuse/issues/3018

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/lib/launcher.js
+++ b/lib/launcher.js
@@ -3,6 +3,7 @@ const { existsSync } = require('fs')
 const fs = require('fs/promises')
 const path = require('path')
 const { info, debug, warn, NRlog } = require('./logging/log')
+const utils = require('./utils')
 
 const MIN_RESTART_TIME = 10000 // 10 seconds
 const MAX_RESTART_COUNT = 5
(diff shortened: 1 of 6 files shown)


hardillb (approved):

knolleary:
I'm not clear why some of the functionality in this PR is required. We already have the ability to use the project nodes in an instance-assigned device. What is different about an app-assigned device that requires all of these changes?

Steve-Mcl:
Hi @knolleary 

> We already have the ability to use the project nodes in an instance-assigned device. What is different about an app-assigned device that requires all of these changes?

Ben and I had a chat about the best way to handle this.  Since we do not yet have the ability to add packages (app device has no settings), we needed a way to get the project-nodes loaded.

The dev-env logic checks to see if there is a `packages/nr-project-nodes` or the NODE_ENV is development & instructs the agent to use the src instead.

> Likewise, we haven't needed nearly as much dev-env specific logic previously - what changed?

There is another place in the code base where the same* dev-env logic is included. The logic in here is pretty much doing the exact same thing (adding to nodes dir) (albeit the code here looks like it is "more" as it was factored out to make it easy to unit test).

 I am not at my computer ATM but I think the same* dev-env logic is in the regular launcher?

---

*same: I additionally test that NODE_ENV is "development"

Steve-Mcl:
> An alternative approach would have been to list them in the default snapshot we provide a device an app-mode.
> 
> https://github.com/FlowFuse/flowfuse/blob/13a7e55072df1359d949109cea5ee3c5e7651709/forge/routes/api/deviceLive.js#L105

Ah, right, so the device agent would only de instructed to install project nodes when the user has updated the core (and the platform is licenced too) 🤔 

Very good point.

I will do some manual testing and determine if we revert this PR  before we ship a device agent update.

Thanks.

PR #225, opened 2024-01-26
Fix affinity cookie parsing

knolleary:
## Description

Spotted a bug I missed in review of the affinity cookie parsing.

### DIFF
--- a/lib/editor/tunnel.js
+++ b/lib/editor/tunnel.js
@@ -70,20 +70,16 @@ class EditorTunnel {
         })
         socket.on('upgrade', (evt) => {
             if (evt.headers && evt.headers['set-cookie']) {
-                const cookies = evt.headers['set-cookie']
-                if (Array.isArray(cookies)) {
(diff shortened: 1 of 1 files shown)


lib/editor/tunnel.js
```diff
@@ -70,20 +70,16 @@ class EditorTunnel {
...
-                } else {
-                    const parts = cookies.split(';')[0].split(['='])
-                    if (parts === 'FFSESSION') {
+                let cookies = evt.headers['set-cookie']
+                if (!Array.isArray(cookies)) {
+                    cookies = [cookies]
+                }
+                cookies.forEach(cookie => {
+                    const parts = cookie.split(';')[0].split(['='])
+                    if (parts[0] === 'FFSESSION') {
```
knolleary, line 79: The core fix was the `[0]` missing when checking the cookie name. I took the time to simplify the 'is it an array or not' logic so I only had to fix this once.

hardillb (approved):

Issue #227, opened 2024-01-30
Possible to have 2 instances of NR running at same time


hardillb:
### Current Behavior

If a restart is triggered while the device agent is already restarting a crashed NR instance then it can start a second instance.

### Expected Behavior

restart to not start a second instance

### Steps To Reproduce

1. force a crash of NR (or a state where a normal kill doesn't stop the process immediately)
2. send a restart command to the device agent

### Environment

- FlowFuse version: 2.0.0
- Node.js version: 18.x
- npm version: 9.x
- Platform/OS: Linux
- Browser: NA

Steve-Mcl:
@hardillb 

Ben, I am exploring the possibility that this _might_ be related to https://github.com/FlowFuse/nr-project-nodes/issues/62 but I am not sure how to recreate:



> * force a crash of NR (or a state where a normal kill doesn't stop the process immediately)

Any pointers on how to achieve this? 

> * send a restart command to the device agent

From where? There are (currently no options from FF core to send a restart to a device).  Do you mean a systemd / service restart?

hardillb:
>         force a crash of NR (or a state where a normal kill doesn't stop the process immediately)
>
> Any pointers on how to achieve this?

A badly written node that doesn't implement a `node.on('close',...)` function and prevents Node-RED from shutting down


>        send a restart command to the device agent
>
>From where? There are (currently no options from FF core to send a restart to a device). Do you mean a systemd / service restart?

No, tag a new snapshot in the forge app, this will tell the device agent to shutdown the NR instance (with a kill -HUP) which then falls foul of previous comment.

Look at the nr-launcher code, it has a test to see if the kill -HUP failed and then sends a kill -9 to ensure the NR instance is dead.

Any code that restarts Node-RED should wait for it to be properly gone before starting a new NR process.

joepavitt:
Can I check the status of this - it's got an attached PR that's been merged, so can it be closed? Or is it just needing to be verified on staging?

Steve-Mcl:
It should have been linked to the PR and therefore would be closed. I'll do it with this comment.

PR #236, opened 2024-02-07
Implement deferred stop like nr-launcher

Steve-Mcl:
## Description

Implement a deferred stop as per nr-launcher

## Related Issue(s)

#227 

Also, potentially the cause of https://github.com/FlowFuse/nr-project-nodes/issues/62

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/lib/launcher.js
+++ b/lib/launcher.js
@@ -2,10 +2,13 @@ const childProcess = require('child_process')
 const { existsSync } = require('fs')
 const fs = require('fs/promises')
 const path = require('path')
-const { info, debug, warn, NRlog } = require('./logging/log')
+const { log, info, debug, warn, NRlog } = require('./logging/log')
 
 const MIN_RESTART_TIME = 10000 // 10 seconds
 const MAX_RESTART_COUNT = 5
+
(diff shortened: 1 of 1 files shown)


knolleary:
@Steve-Mcl what's the connection to the Project Nodes issue?

For the Project Nodes issue, Rob was using the Modified Nodes deploy type to make changes. I don't believe restarting Node-RED had anything to do with it.

knolleary:
Sorry - just caught up on the comments in the linked issue. I'm not certain. Both instances would be trying to listen on the same port and one will fail to start, so I don't see how that results in two copies of the flows running in parallel.

hardillb:
This is something that required fixing as it was possible to fail to start due to the old version still being running, but as you say I don't think is the cause of Rob's issue

hardillb:
I'm going to merge this, but as mentioned I don't think it's the cause of the nr-project-nodes problem

Steve-Mcl:
@knolleary @hardillb right you are. Thanks for the cycles.

PR #239, opened 2024-03-10
Add support for device actions

Steve-Mcl:
## Description

* Adds new command route to `mqtt.js`: `"action"`
  * Supports "start", "suspend", "restart"
* Adds `targetState` property to the project (persisted to the project file to ensure reboots to correct target state)
   * targetState supports only "running" and "suspended" at this time
* Converted hardcoded `"state"` strings into a an enum (was getting lost with all the supported states) $ added helper functions for verifying state
* To support timely and more accurate state feedback to FF, `checkin` is now called when state changes inside the launcher (important now we have transitional state info reporting on FF and necessary for good UX on the Actions menu items availability)
* Attempts to mitigate issue when agent is busy installing modules and the launcher is stopped by awaiting the install process for a few seconds grace before forcefully stopping. This was done as I managed cause a lock ups by ill timed restarts and suspends during testing.
  * NOTE: This was a balance struck on making the device "do what it is told" vs "leaving it "stuck installing" or "stuck starting" etc. 
  * We may want to revisit and simply inhibit actions altogether while installation is occurring.

### Tests added
```
  Agent
    actions
      ✔ suspends Node-RED
      ✔ reloads target state as suspended
      ✔ reloads target state as running
      ✔ restarts Node-RED
      ✔ starts Node-RED after suspend
    getState
      ✔ returns state when suspended

  MQTT Comms
    ✔ suspend action calls agent.suspendNR and checks in (573ms)
    ✔ start action calls agent.startNR and checks in (567ms)
    ✔ restart action calls agent.restartNR and checks in (576ms)
    ✔ Invalid action responds with error (568ms)

  State validation
    ✔ isValidState should return true for valid states
    ✔ isValidState should return false for invalid states
    ✔ isTargetState should return true for valid target states
    ✔ isTargetState should return false for invalid target states
    ✔ isTransitionState should return true for valid transition states
    ✔ isTransitionState should return false for invalid transition states
```

## Related Issue(s)

https://github.com/FlowFuse/flowfuse/issues/3292

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/lib/launcher.js
+++ b/lib/launcher.js
@@ -406,7 +406,9 @@ class Launcher {
         })
 
         this.proc.on('exit', async (code, signal) => {
-            const expected = ['shutdown', 'restart', States.UPDATING, States.SUSPENDED].includes(this.stopReason)
+            // determine if Node-RED exited for an expected reason
+            // if yes, don't restart it since it was specifically stopped (e.g. not crashed)
+            const expected = ['shutdown', States.RESTARTING, States.UPDATING, States.SUSPENDED].includes(this.stopReason)
             if (expected) {
                 this.state = States.STOPPED // assume stopped
             } else {
--- a/lib/launcher.js
+++ b/lib/launcher.js
@@ -4,6 +4,7 @@ const fs = require('fs/promises')
 const path = require('path')
 const { log, info, debug, warn, NRlog } = require('./logging/log')
 const { copyDir } = require('./utils')
+const { States } = require('./states')
 
 const MIN_RESTART_TIME = 10000 // 10 seconds
 const MAX_RESTART_COUNT = 5
@@ -22,8 +23,8 @@ const packageJSONTemplate = {
 }
 
 class Launcher {
-    constructor (config, application, project, snapshot, settings, mode) {
-        this.config = config
+    constructor (agent, application, project, snapshot, settings, mode) {
+        this.config = agent?.config
         this.application = application
         this.project = project
         this.snapshot = snapshot
@@ -31,7 +32,12 @@ class Launcher {
         this.mode = mode
         this.restartCount = 0
         this.startTime = []
-        this.state = 'stopped'
+        this.state = States.STOPPED
+        this.stopReason = ''
+        this.installProcess = null
+        this.deferredStop = null
+        /** @type {import('./agent.js').Agent */
+        this.agent = agent
 
         this.auditLogURL = `${this.config.forgeURL}/logging/device/${this.config.deviceId}/audit`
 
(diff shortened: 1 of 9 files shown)


lib/launcher.js
```diff
@@ -439,8 +458,22 @@ class Launcher {
...
+            // to avoid corruption (NPM can leave temporary directories preventing future installs)
+            // We should wait for the install to finish before stopping
+            // give it a few seconds to finish
+            const timeout = new Promise(resolve => setTimeout(resolve, 10000))
+            await Promise.race([this.installProcess, timeout])
+            // now proceed with stopping, regardless of whether the install finished
+        }
+        let finalState = States.STOPPED
+        this.stopReason = reason || 'shutdown'
+        info('Stopping Node-RED. Reason: ' + reason)
```
Steve-Mcl, line 473:
```suggestion
        info('Stopping Node-RED. Reason: ' + this.reason)
```

@hardillb, this fixes the `undefined` log entry in this branch

hardillb (commented):
Runs and works as expected with the changes to the forge app.

Appears to work properly with staging  (no changes to forge app yet)

One thing about debug printing `[Object]` not actually in this diff, but flagged.

lib/launcher.js
```diff
@@ -439,8 +458,22 @@ class Launcher {
...
+            const timeout = new Promise(resolve => setTimeout(resolve, 10000))
+            await Promise.race([this.installProcess, timeout])
+            // now proceed with stopping, regardless of whether the install finished
+        }
+        let finalState = States.STOPPED
+        this.stopReason = reason || 'shutdown'
+        info('Stopping Node-RED. Reason: ' + this.reason)
+        if (reason === States.SUSPENDED) {
+            finalState = States.SUSPENDED
+        }
```
Steve-Mcl, line 476:
```suggestion
        this.stopReason = reason || 'shutdown'
        info('Stopping Node-RED. Reason: ' + this.stopReason)
        if (this.stopReason === States.SUSPENDED) {
            finalState = States.SUSPENDED
        }
```

@hardillb apologies, i messed up.  fixed now.

hardillb (approved):

knolleary:
No structural changes being made in the forge app around **how** it will send commands to the device agent. We are only shipping the `restart` command initially as there are unresolved questions on the semantics of `suspend` (and thus `start` is also unnecessary).

@Steve-Mcl I think we're okay to merge this as-is?

Steve-Mcl:
> No structural changes being made in the forge app around **how** it will send commands to the device agent. We are only shipping the `restart` command initially as there are unresolved questions on the semantics of `suspend` (and thus `start` is also unnecessary).
> 
> @Steve-Mcl I think we're okay to merge this as-is?

Yeah. Ben and I have tested against local and staging.

PR #293, opened 2024-07-17
Add heartbeat to Editor WebSocket Tunnel

### LINKED ISSUE #288: Add heartbeat to editor tunnel websocket
### Description

If the device editor has the editor tunnel connected, but there is no active editor open, the websocket goes idle and gets disconnected by the platform regularly. This fills the logs with connection closed/reconnecting log messages - which can make the platform appear unstable to the user.

It would be better to have a heartbeat keeping the connection alive.

Will need to be done in a way that doesn't upset the platform when this new heartbeat packet arrives from an updated device agent.

### Epic/Story

_No response_

hardillb:
fixes #288

## Description

Sends a Websocket Ping every ~15~ 40 seconds to ensure the Editor Tunnel WebSocket stays open

## Related Issue(s)

#288

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/lib/editor/tunnel.js
+++ b/lib/editor/tunnel.js
@@ -248,6 +248,7 @@ class EditorTunnel {
                     })
                 }
             })
+            clearInterval(this.pingInterval)
             this.pingInterval = setInterval(() => {
                 debug('Sending WS editor tunnel ping')
                 socket.ping()
--- a/lib/editor/tunnel.js
+++ b/lib/editor/tunnel.js
@@ -248,6 +248,10 @@ class EditorTunnel {
                     })
                 }
             })
+            this.pingInterval = setInterval(() => {
+                debug('Sending WS editor tunnel ping')
+                socket.ping()
+            }, 15000)
         }
         socket.on('close', async (code, reason) => {
             if (Buffer.isBuffer(reason)) {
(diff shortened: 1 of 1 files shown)


lib/editor/tunnel.js
```diff
@@ -248,6 +248,11 @@ class EditorTunnel {
                     })
                 }
             })
+            clearInterval(this.pingInterval)
+            this.pingInterval = setInterval(() => {
+                debug('Sending WS editor tunnel ping')
+                socket.ping()
+            }, 15000)
```
knolleary, line 255:
```suggestion
            }, 40000)
```

We see the timeouts at the 60s mark - so I think we can afford to reduce the ping frequency a bit here - will halve the traffic it generates otherwise :)

hardillb: They about as small as we can send, but good point. I was just too impatient to wait that long while testing....

lib/editor/tunnel.js
```diff
@@ -248,6 +248,11 @@ class EditorTunnel {
                     })
                 }
             })
+            clearInterval(this.pingInterval)
+            this.pingInterval = setInterval(() => {
+                debug('Sending WS editor tunnel ping')
+                socket.ping()
```
knolleary, line 254: Suggest adding a guard here just in case the socket is disconnected when this fires (and the clearTimeouts haven't happen yet) - we do not want an `UncaughtException` killing us here.
```suggestion
                try {
                    socket.ping()
                } catch (err) { }
```

Issue #263, opened 2024-05-24
Remove Node 14 support in Device Agent 3.x due to 'inflight' deprecation


knolleary:
As of May 23rd, the `inflight` module has been deprecated. This means users now see the following warning when they install the module:

> npm WARN deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.

The module is deep in the `mqtt` module dependency tree (see below), but more importantly, it is not in the functional code path of the device agent. So our end users are in no way impacted.

However, that install warning is alarming and has already triggered support questions within a day of the deprecation. Ideally we need to remove it from our dependency tree. However, that brings some complication.

The module is in the dependency tree of `mqtt@4.x`. It is *not* in the dependency tree of `mqtt@5.x` - however, 5.x drops support for Node 14 - something we want to try to keep for some of the device manufacturers in the ecosystem. I think it is unlikely that the `mqtt` project will accept a patch for the 4.x stream - they haven't done any maintenance releases on it since 5.x was released two years ago.

This was a somewhat inevitable outcome whilst trying to keep support for such an old version of node.js.

We have two options:

1. Do nothing other than document it. No-one reads the docs, but at least there will be an answer to be found when searching.
2. Update mqtt to 5.x and drop support for Node 14.

Moving forward, option 2 feels the only way forward. Node 14 is very old and this is only going to be the first of many such cases where fixes are not made available in older versions of our dependencies.

I propose we bump the Device Agent to 3.x. We can still support Device Agent 2.x on Node 14, but will be in maintenance mode (unless there is a specific business need to backport something).

knolleary:
We have had enquires regarding our plan to eliminate the deprecated dependencies. Whilst it still holds true they are not in any functional code path used by the Device Agent, the presence of the deprecation warnings is being flagged as a blocker to deploy the module for some users.

Adding to the 2.8 scheduled to get resolved.

Steve-Mcl:
Reviewing https://github.com/FlowFuse/device-agent/pull/306 it hit me that project nodes will require a similar fix and bump. 

dependency: https://github.com/FlowFuse/nr-project-nodes/blob/2344397452a694e9e6e0cd8c488083fb8a6cd5f6/package.json#L46
[...]

Issue #267, opened 2024-05-30
Multiple nic (communication) card causes disconnections for gateway type devices


gdziuba:
### Current Behavior

When a device with multiple nic cards runs the device agent.  It seems as if it causes network issues on the device and the device then starts to try to switch between communications.  

Device issues can be elevated when one of the communication cards is disabled.

We have been working with robustel and have access to a device where we can reduplicate.  Please reach to go @gdziuba for details on how to connect to the device.

### Expected Behavior

Device agent to maintain connection to the platform.

### Steps To Reproduce

In this case we leverage an EG5120 from robustel.  Running device agent with a sim card and wifi will cause the device to start disconnecting and reconnecting.

### Environment

- FlowFuse version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

### Linked Customers

- Customer name and/or link to HubSpot contact
This customer seems to be having similar issues with a different product.

https://app-eu1.hubspot.com/contacts/26586079/record/0-2/11347800779

gdziuba:
@joepavitt, @knolleary.  This is the issue that I raised concern for in the #dept-engineering channel.

knolleary:
@gdziuba  It would be useful to have more context on this - we don't have much to go on here. It feels like this is going to be a very device-specific thing and we haven't had any other reports of this type of behaviour.

The EG5120 was mentioned in a support ticket earlier this week regarding IPv6 connectivity. I wonder if that is related at all here.

gdziuba:
@knolleary That is a fair statement.  

The customer was incorrect above and will be updated: https://app-eu1.hubspot.com/contacts/26586079/record/0-2/7790198479

Customer notes:

Compulab IoT gateway
Debian
IOT-GATE-IMX8PLUS - Industrial IoT Gateway
Cellular Communication
Bandwidth issues
Agent Version is 2.3.2 (Updated to 2.4.1 and still had issues)
Had to restart the service
Not Docker
TCP and UDP issues on device agents


Reduplicate the issue:

1. Clean device
2. Install device agent
3. Start
4. Pulls Node-RED down
5. Deploy a flow
6.  Try to reopen editor and it doesn't connect
7. Restart device agent service
8. Typically, opening editor is fine until another deploy.

Resolution for EG5120
1.  Connect to Ethernet
2. Disable other communication cards (wifi, cellular)

hardillb:
We will need to know how the network is configured on both interfaces and what routes are configured with what weights

[...]

PR #306, opened 2024-08-13
Upgrade to latest MQTT.js

hardillb:
part of #263
## Description

Update mqtt.js and engines section of package.json.

Also had to change the client.connect()` call slightly to get it to work properly.

## Related Issue(s)

#263

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/lib/mqtt.js
+++ b/lib/mqtt.js
@@ -67,7 +67,13 @@ class MQTTClient {
                 agent: getWSProxyAgent(this.config.brokerURL)
             }
         }
-        this.client = mqtt.connect(brokerURL.href, this.brokerConfig)
+        const connectOpts = {
+            protocol: brokerURL.protocol.replace(/:$/, ''),
+            host: brokerURL.hostname,
+            port: brokerURL.port
+        }
+        const opts = Object.assign({}, connectOpts, this.brokerConfig)
+        this.client = mqtt.connect(opts)
 
         this.client.on('connect', () => {
             info('MQTT connected')
--- a/lib/mqtt.js
+++ b/lib/mqtt.js
@@ -67,7 +67,7 @@ class MQTTClient {
                 agent: getWSProxyAgent(this.config.brokerURL)
             }
         }
-        this.client = mqtt.connect(brokerURL, this.brokerConfig)
+        this.client = mqtt.connect(brokerURL.href, this.brokerConfig)
 
         this.client.on('connect', () => {
             info('MQTT connected')
--- a/lib/mqtt.js
+++ b/lib/mqtt.js
@@ -50,17 +50,7 @@ class MQTTClient {
     }
 
     start () {
-        // PROBLEM: ipv6 ws addresses cannot connect
-        // INFO: Calling mqtt.connect('http://[::1]:8883') fails with error  ERR_INVALID_URL
-        // INFO: Calling mqtt.connect(new URL('http://[::1]:8883')) fails because `connect` only accepts a `string` or `url.parse` object
-        // INFO: Calling mqtt.connect(url.parse('http://[::1]:8883') fails because unlike new URL, url.parse drops the square brackets off hostname
-        //       (mqtt.js disassembles and reassembles the url using hostname + port so `ws://[::1]:8883` becomes `ws://::1:8883`)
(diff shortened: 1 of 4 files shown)


hardillb:
These are the changes I needed to make to get #267 to work properly

lib/mqtt.js
```diff
@@ -66,7 +66,7 @@ class MQTTClient {
                 agent: getWSProxyAgent(this.config.brokerURL)
             }
         }
-        this.client = mqtt.connect(brokerURL, this.brokerConfig)
+        this.client = mqtt.connect(brokerURL.href, this.brokerConfig)
```
Steve-Mcl, line 69: As far as I can see, this will cause the ERR_INVALID_URL error mentioned in the code above.

TLDR;
Using `brokerURL.href` results in sending the original URL (e.g. 'ws://[::1]:8883/'`) which gets manually deconstructed and reconstructed in the bowels of the MQTT lib as 'ws://::1:8883/'` resulting in the original problem.


DETAILS:

REF: https://github.com/mqttjs/MQTT.js/issues/1569

I have looked at the src for V5 and they [still use](https://github.com/mqttjs/MQTT.js/blob/accd78e38aa82c8cc1ea04029e56494276776c87/lib/connect/index.js#L64) the depreciated `url.parse` to deconstruct the URL then reconstruct it manually which is what out workaround was fixing.


The only thing I am not 100% certain of is if the `import url` at the top of the v5 code causes the code to use the old depreciated `url.parse` or the new `new URL` methods after the ts is compiled.  If it is the latter, the problem goes away.



To confirm this I will need to pull this branch and do local testing with IPV6

Steve-Mcl: Unfortunately, this does indeed cause an error when a IPV6 IP URL is used.

Looking through the `MQTT.js` src, i cannot see how passing a string vs passing the URL object makes any difference to reconnect options.

If I have missed something there may be an alternative route. If we must send a string URL then there is an opportunity to intercept and update the final connection URL by setting the connect opts function `transformWsUrl` 

NOTE: I have raised a PR against the linked issue on `MQTT.js`

hardillb: Passing the object was leading to it trying to  connect to localhost in the device agent even when the URL pointed to staging.

Steve-Mcl:
Looks good ben but [the whole workaround code](https://github.com/FlowFuse/device-agent/pull/306/files#diff-22d60ae947b2cc2a1180672627a138f507749c3f98792237cd599d075ec38998R52-R62) can go in the bin now.

I would add a suggestion to save you messing on but its out of the touched code and GH wont let me.

```
        // PROBLEM: ipv6 ws addresses cannot connect
        // INFO: Calling mqtt.connect('http://[::1]:8883') fails with error  ERR_INVALID_URL
        // INFO: Calling mqtt.connect(new URL('http://[::1]:8883')) fails because `connect` only accepts a `string` or `url.parse` object
        // INFO: Calling mqtt.connect(url.parse('http://[::1]:8883') fails because unlike new URL, url.parse drops the square brackets off hostname
        //       (mqtt.js disassembles and reassembles the url using hostname + port so `ws://[::1]:8883` becomes `ws://::1:8883`)
        // INFO: WS src code uses `new URL` so when `mqtt.js` passes the reassembled IP `http://::1:8883`, it fails with error ERR_INVALID_URL
        // SEE: https://github.com/mqttjs/MQTT.js/issues/1569
        // eslint-disable-next-line n/no-deprecated-api
        const brokerURL = url.parse(this.config.brokerURL)
        const _url = new URL(this.config.brokerURL)
        brokerURL.hostname = _url.hostname
```

simply becomes 

```
        const brokerURL = new URL(this.config.brokerURL)
```

Steve-Mcl (approved):
Approved...

MQTT JS v5.10 is now released - do we want to bump to `latest` or leave as is here?

Will let you decide to bump or merge :)

hardillb:
Not merging until we decide the release strategy for this, as I think  we need to tag a NodeJS v14 support branch

knolleary (approved):

hardillb:
Just a moment, rebasing to pick up the package-lock.json

PR #313, opened 2024-09-03
Use port from config file

### LINKED ISSUE #312: Port option from device.yml Not Being Applied
### Current Behavior

**Problem**: Changing the port via the device.yml file is not working.

**Cause**: In the config function within **device-agent/lib/config.js**, the return value is constructed by merging parsedOptions.deviceConfig and options using the spread operator.

Since options always includes the default port 1880 (from device-agent/lib/cli/args.js), even without the -p argument, the port setting from parsedOptions.deviceConfig is never applied because the last spread operator takes precedence.

**Possible Solution**: To fix this issue, ensure that parsedOptions.deviceConfig takes precedence over options by adjusting the order in which the spread operator is applied.

changing this:
`return {
        version,
        ...parsedOptions.deviceConfig,
        ...options
    }
`
to this:
`return {
        version,
        ...options,
        ...parsedOptions.deviceConfig
    }`

makes the port setting in device.yml the priority setting

### Expected Behavior

Adding port option to device.yml should change the port

### Steps To Reproduce

Add port to device.yml for example 2222
start device agent without port command-line argument
🠊 node-RED still runs at 1880

### Environment

- FlowFuse version: Flowfuse Cloud
- Node.js version: 20.17.0
- npm version: 10.8.2
- Platform/OS: Windows Server 2022
- Browser: Edge

### Linked Customers

_No response_

hardillb:
fixes #312

## Description

Default options overwrite port number in config file

## Related Issue(s)

#312 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/index.js
+++ b/index.js
@@ -207,7 +207,10 @@ Please ensure the parent directory is writable, or set a different path with -d`
         return {
             AgentManager,
             webServer,
-            options
+            options: {
+                ...ConfigLoader.defaults,
+                ...options
+            }
         }
     }
     return null
--- a/lib/cli/args.js
+++ b/lib/cli/args.js
@@ -32,7 +32,6 @@ module.exports = [
         name: 'port',
         alias: 'p',
         type: Number,
-        defaultValue: 1880,
         typeLabel: '{underline number}',
         group: 'main'
     },
@@ -67,7 +66,6 @@ module.exports = [
         name: 'ui',
         description: 'Start the Web UI Server (optional, does not run by default)',
         type: Boolean,
-        defaultValue: false,
         alias: 'w',
         group: 'ui'
     },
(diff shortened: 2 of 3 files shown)


knolleary:
I don't think this is the approach we want to take as it will mean explicitly set command-line args do not override the settings file.

The core issue is the fact that if no port option is provided on the command line, the code defaults as if 1880 had been set on the command-line - and overrides the file.

I think the order of precedence should be:

 1. Command Line
 2. Settings file
 3. Default value

That will require a bit more to achieve than just reordering the spread operations

hardillb:
OK,

So I think if we 
- remove the default options from the default options from the cli config
- add a default options object to the start of the spread
- swap the order back to cli parse options last in the spread

e.g.

```
const defaults = {
  port: 1880.
  ...
}

return {
        version,
        ...defaults,
        ...parsedOptions.deviceConfig
        ...options,
    }
```

knolleary (approved):

PR #315, opened 2024-09-09
Randomise reconnect timing

### LINKED ISSUE #314: Device Agent reconnect delay

Steve-Mcl:
closes #314

## Description

To alleviate stresses on broker, a variation in the reconnect timing is introduced

The exact value may wish to be tweaked but I went with upto 2 seconds sooner and no more than 10s later than the original 15s fixed time

## Related Issue(s)

#314

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/.mocharc.js
+++ b/.mocharc.js
@@ -1,0 +1,1 @@
+process.env.NODE_ENV = 'test'
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -9,6 +9,7 @@ const Launcher = require('./launcher.js')
 const { info, warn, debug } = require('./logging/log')
 const utils = require('./utils.js')
 const { States, isTargetState, isValidState } = require('./states')
+const MQTT_CONNECT_DELAY_MAX = process.env.NODE_ENV === 'test' ? 25 : 5000
 
 const PROJECT_FILE = 'flowforge-project.json'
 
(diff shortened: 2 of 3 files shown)


knolleary:
I think we should also include a similar randomised delay on startup. Perhaps not as extreme as 15secs, but between 2-10 secs for eg.

Steve-Mcl:
> I think we should also include a similar randomised delay on startup. Perhaps not as extreme as 15secs, but between 2-10 secs for eg.

As the mqtt.connect call is synchronous and following events rely on their being a client object created, the straightforward solution to this is to add a random await before [the call to `agent.start`](https://github.com/FlowFuse/device-agent/blob/8b1c63f1064388bc95cad35a3f2cf44e2bbb8384/lib/AgentManager.js#L99)



There is only 1 reference to the `agent.start` function (and it is already an async fn), so a simple awaited delay:
```
await new Promise(r => setTimeout(r, randomInt(20, 5000)));
```
before [line 138](https://github.com/FlowFuse/device-agent/blob/8b1c63f1064388bc95cad35a3f2cf44e2bbb8384/lib/agent.js#L138):
```
this.mqttClient.start()
```
would be feasible.

Steve-Mcl:
> a simple awaited delay:
> 
> ```
> await new Promise(r => setTimeout(r, randomInt(20, 5000)));
> ```
> 
> before [line 138](https://github.com/FlowFuse/device-agent/blob/8b1c63f1064388bc95cad35a3f2cf44e2bbb8384/lib/agent.js#L138):
> 
> ```
> this.mqttClient.start()
> ```
> 
> would be feasible.

Added in commit [add randomised start call delay](https://github.com/FlowFuse/device-agent/pull/315/commits/22ae18b17655f12123865fbb0fa91a1b9801cb36)


As this affects the start up time, I have kept is small enough to be less bothersome (20ms ~ 5s)

knolleary:
@Steve-Mcl looks good - but a failing lint check in the tests

Steve-Mcl:
> @Steve-Mcl looks good - but a failing lint check in the tests

🤦‍♂️ Thats what I get for doing it in GH 

fixed in [691eae7](https://github.com/FlowFuse/device-agent/pull/315/commits/691eae7327bc129aa43a95c13d0aa781771c787a) 🤞

Steve-Mcl:
and of course, tests now timeout 

```
Timeout of 2000ms exceeded
```

fix incoming.

knolleary (approved):

PR #316, opened 2024-09-13
Bump minimum nodejs version

### LINKED ISSUE #263: Remove Node 14 support in Device Agent 3.x due to 'inflight' deprecation

hardillb:
fixes #263 
## Description

Updates the minimum NodeJS version to 18

## Related Issue(s)

#263 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


hardillb:
Once merged #306 can also be merged

knolleary (approved):

PR #324, opened 2024-10-24
Bump for 3.0.0

Steve-Mcl:
## Description

Bump for 3.0.0


Steve-Mcl:
@knolleary I did not call out the dropping of Node 14/16 in the CHANGELOG.md notes - should we explicitly point that out in the 3.0.0 release notes to make it prominent?

knolleary:
Yes, we should. This also needs including the release blog post and docs. Lets have a chat about the plan for this.

PR #325, opened 2024-11-08
Log reason Provisioning failed

hardillb:
## Description

Customer hit this today and neither side logs why the provisioning failed. This will at least hint

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/lib/AgentManager.js
+++ b/lib/AgentManager.js
@@ -208,6 +208,7 @@ class AgentManager {
             provisioningOK = true
         } catch (err) {
             warn(`Problem encountered during provisioning: ${err.toString()}`)
+            warn(`Reason: ${err.response?.body}`)
             provisioningOK = false
         }
 


lib/AgentManager.js
```diff
@@ -208,6 +208,7 @@ class AgentManager {
             provisioningOK = true
         } catch (err) {
             warn(`Problem encountered during provisioning: ${err.toString()}`)
+            warn(`Reason: ${err.response?.body}`)
```
Steve-Mcl, line 211: if `err.response` is empty, this will just print "Reason:" - is that desired?

Maybe better to conditionally print the reason?:

```suggestion
            if (err.response?.body) {
                warn(`Reason: ${err.response.body}`)
            }
```

Happy either way if that was the intention.

hardillb: It should never be empty, and reporting the empty state if it exists is helpful

Steve-Mcl: "Should" is the operatave word ;) 

The point is if it is, the `?.` will permit the `warn` to print "Reason:".  Similarly, if `body` is empty, it will print "Reason:"

Perhaps better would be:

```suggestion
            warn(`Reason: ${err.response?.body || 'unknown'}`)
```

or similar to avoid the edge case?

Steve-Mcl (approved):

PR #327, opened 2024-11-18
Send nr version

hardillb:
## Description

Has the the device agent report the currently running Node-RED version as part of it's regular check-in

This for use with the Device Editor Proxy.

## Related Issue(s)

FlowFuse/flowfuse#3414

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -261,7 +261,7 @@ class Agent {
             agentVersion: this.config.version,
             licensed: this.config.licensed
         }
-        if (this.launcher) {
+        if (this.launcher?.readPackage) {
             const { modules } = await this.launcher.readPackage()
             if (module['node-red'] !== 'latest') {
                 state.nodeRedVersion = modules['node-red']
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -263,7 +263,7 @@ class Agent {
         }
         if (this.launcher) {
             const { modules } = await this.launcher.readPackage()
-            if (module['node-red'] != 'latest') {
+            if (module['node-red'] !== 'latest') {
                 state.nodeRedVersion = modules['node-red']
             } else {
                 const nrPackPath = path.join(this.launcher.projectDir, 'node_modules/node-red/package.json')
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -268,7 +268,7 @@ class Agent {
                 state.nodeRedVersion = modules['node-red']
             } else {
                 const nrPackPath = path.join(this.launcher.projectDir, 'node_modules/node-red/package.json')
-                const content =  readFileSync(nrPackPath)
+                const content = readFileSync(nrPackPath)
                 const packJSON = JSON.parse(content)
                 state.nodeRedVersion = packJSON.version
             }
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -264,7 +264,7 @@ class Agent {
         }
         if (this.launcher?.readPackage) {
             const { modules } = this.launcher.readPackage()
-            if (module['node-red'] !== 'latest') {
+            if (modules['node-red'] !== 'latest') {
                 state.nodeRedVersion = modules['node-red']
             } else {
                 const nrPackPath = path.join(this.launcher.projectDir, 'node_modules/node-red/package.json')
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -265,7 +265,7 @@ class Agent {
         }
         if (this.launcher?.readPackage) {
             const { modules } = this.launcher.readPackage()
-            if (!!semver.valid(modules['node-red'])) {
+            if (semver.valid(modules['node-red']) !== null) {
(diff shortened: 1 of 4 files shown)


Steve-Mcl (changes requested):
Ben, I can see a few other places in the code base where `getAgent` is called but without the async. And where they are, they have sync function signatures meaning the proliferation of the `async` `await` stuff has tentacles that reach even further out & I am unsure of the sideeffects in production.

image





TBH, I would be tempted to undo all of this and simply modify `launcher.readPackage()` to use `readFileSync` from `fs`.

There are only around ~5 calls to `readPackage` that would simply need to lose the `await` - then there would be no need to make `getState` async and no need to follow its code branches and parent callers adding await to everything it touches.

lib/mqtt.js
```diff
@@ -195,7 +195,7 @@ class MQTTClient {
         }
     }
 
-    checkIn () {
+    async checkIn () {
```
Steve-Mcl, line 198:
```suggestion
    checkIn () {
```

Missed one ;)

lib/agent.js
```diff
@@ -261,6 +262,17 @@ class Agent {
             agentVersion: this.config.version,
             licensed: this.config.licensed
         }
+        if (this.launcher?.readPackage) {
+            const { modules } = this.launcher.readPackage()
+            if (module['node-red'] !== 'latest') {
```
Steve-Mcl, line 267:
```suggestion
            if (modules['node-red'] !== 'latest') {
```

module is the actual `module` -  i suspect this should be `modules`

lib/agent.js
```diff
@@ -261,6 +262,17 @@ class Agent {
...
+                state.nodeRedVersion = modules['node-red']
+            } else {
+                const nrPackPath = path.join(this.launcher.projectDir, 'node_modules/node-red/package.json')
+                const content = readFileSync(nrPackPath)
+                const packJSON = JSON.parse(content)
+                state.nodeRedVersion = packJSON.version
```
Steve-Mcl, line 273: While I am certain this will work almost every time, I am always nervous about reading files AND parsing JSON without a try..catch.

Since this is mostly informational type stuff, should we wrap this (or the whole section) in a try..catch)?

lib/agent.js
Steve-Mcl, line 267: Since the goal is to get a semver, should we be testing `modules['node-red']` is a semver?

e.g. something like: `semver.valid(modules['node-red'])` instead?

Steve-Mcl (changes requested):
Yeah, much cleaner without all the proliferating async await.

However, there is an issue with `module` vs `modules` and a couple of suggestions.

Lastly, do we thing a unit test is possible? 

there is already tests in `test/unit/lib/agent_spec.js` around line 493 for `getState` - even just the presence of the property being a valid semver

lib/agent.js
hardillb: If the node-red package.json doesn't parse it won't have been installed or the filesystem is broken and their is no way to carry on...

lib/agent.js
hardillb: I was trying to not add any more dependencies

hardillb:
> Lastly, do we thing a unit test is possible?
> 
> there is already tests in test/unit/lib/agent_spec.js around line 493 for getState - even just the presence of the property being a valid semver

I don't see how without having the test actually install Node-RED

lib/agent.js
Steve-Mcl: semver is already a dependency: https://github.com/FlowFuse/device-agent/blob/de9bd72af2ba5ca8311adb9eb97ca15c4a4c4f70/package.json#L45

Steve-Mcl:
> I don't see how without having the test actually install Node-RED.
> 
> I already had to patch the code because the tests are injecting a JSON object rather than a real launcher object

Fair enough Ben - just thought to ask.

Steve-Mcl:
can we look at using `semver.valid` since semver is already a dependency?

Without this, I can foresee a future ticket where the `nodeRedVersion` is  `*` or  `dev` :)

lib/agent.js
```diff
@@ -261,6 +263,21 @@ class Agent {
             agentVersion: this.config.version,
             licensed: this.config.licensed
         }
+        if (this.launcher?.readPackage) {
+            const { modules } = this.launcher.readPackage()
+            if (!semver.valid(modules['node-red'])) {
```
Steve-Mcl, line 268: should this be inverted Ben?

Steve-Mcl:
```suggestion
            if (semver.valid(modules['node-red'])) {
```

I think the `!` is a left over

hardillb: No, it should have had 2 because the output is either null or a string, but the lint rules don't like that now.

Steve-Mcl (approved):

PR #332, opened 2024-12-17
Apply custom message in login dialog on direct access

knolleary:
Fixes https://github.com/FlowFuse/node-red/issues/25

This PR removes the username/password fields from the login dialog when the editor is accessed directly. We get support requests asking what credentials to use - however we only support accessing the editor via the device agent proxy.

With NR 4.0.6 (to be released imminently), I've added the ability to insert a custom message/button on the login dialog when in this state. Older versions of NR will show a blank dialog.

image

Note: spotted the `dashbaord` typo when I took this screenshot for the PR... it will be fixed momentarily.

### DIFF
--- a/lib/template/template-settings.js
+++ b/lib/template/template-settings.js
@@ -42,7 +42,6 @@ let got
 let agentApplied
 
 const auth = {
-    type: 'credentials', // the type of the auth
     tokenHeader: 'x-access-token', // the header where node-red expects to find the token
     tokens: async function (token) {
         const [prefix, deviceId] = (token + '').split('_')
@@ -94,12 +93,6 @@ const auth = {
         } catch (err) {
             console.log('error getting new token', err)
         }
-    },
-    users: async function (username) {
-        return null
-    },
-    authenticate: async function (username, password) {
-        return null
     }
 }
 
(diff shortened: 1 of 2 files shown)


lib/template/template-settings.js
```diff
@@ -169,6 +162,15 @@ const runtimeSettings = {
     [themeName]: { ...themeSettings },
     editorTheme: { ...editorTheme }
 }
+runtimeSettings.editorTheme.login = {
+    message: 'Access to the editor is managed through the FlowFuse platform'
+}
+if (themeSettings.projectURL) {
+    runtimeSettings.editorTheme.login.button = {
+        url: themeSettings.projectURL,
+        label: 'Open FlowFuse Dashbaord'
```
knolleary, line 171:
```suggestion
        label: 'Open FlowFuse Dashboard'
```

Steve-Mcl:
@knolleary not sure if you noticed test fail following removal of `type`?

knolleary:
@Steve-Mcl test fix pushed

Steve-Mcl (approved):
Tested/working/approved

This is how it looks on Win + Chrome

image

Aesthetically, I do wish the string didn't wrap on the last word (but at the same time I cant think of a better way to phrase it)

Steve-Mcl:
> Aesthetically, I do wish the string didn't wrap on the last word (but at the same time I cant think of a better way to phrase it)

Perhaps `'Access to the editor is managed via FlowFuse'` would be a suitable (shorter, non wrapping) alternative?

lib/template/template-settings.js
```diff
@@ -169,6 +162,15 @@ const runtimeSettings = {
     [themeName]: { ...themeSettings },
     editorTheme: { ...editorTheme }
 }
+runtimeSettings.editorTheme.login = {
+    message: 'Access to the editor is managed through the FlowFuse platform'
```
knolleary, line 166:
```suggestion
    message: 'Access the editor through the FlowFuse platform'
```

knolleary:
Have updated the message to `Access the editor through the FlowFuse platform`

PR #341, opened 2025-01-30
Add support for httpNodeAuth

knolleary:
Part of https://github.com/FlowFuse/flowfuse/issues/5014
## Description
This PR adds support for FF User authentication for the HTTP Node/Dashboard routes.

 - It adds the FF Auth plugin. For now, its a copy/paste of the plugin from the `nr-launcher` repo - was easier than trying to repackage.
 - It maintains support for setting `httpNodeAuth` in the `device.yml` file - that takes precedence over whatever is configured via the platform
 - Bumps device agent to v3.1.0 - this is the feature used on the platform-side PR to detect if these options can be applied to the device.
 - Platform-side PR: https://github.com/FlowFuse/flowfuse/pull/5068

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,2 @@
-node_modules/
+/node_modules/
 var/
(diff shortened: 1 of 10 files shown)


knolleary:
@Steve-Mcl I've tagged you for a review here. We can discuss details on Monday morning, but would appreciate if you're able to spare an hour on this.

knolleary:
We can safely merge this one - it is inert until the user enables FF Auth via #5068

PR #348, opened 2025-02-12
Reload device settings when restarting in dev mode

knolleary:
Closes https://github.com/FlowFuse/flowfuse/issues/5069

When in developer mode, if the user triggers a restart via the UI, or if starting up the agent from scratch, trigger a reload of settings from the platform.

This ensures that any env vars/settings changes get picked up in an expected way. Previously this would require taking the device out of developer mode - which then resets the flows etc.

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -185,7 +185,7 @@ class Agent {
             this.launcher = undefined
         }
         await this.updateTargetState(States.RUNNING)
-        await this.setState({ targetState: States.RUNNING })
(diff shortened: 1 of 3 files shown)


knolleary:
To verify this behaviour:

1. Connect device, put in dev mode
2. Deploy flow to report (for example) value of an env var
3. Edit env var in platform UI
4. Trigger restart action in platform UI
5. Verify the flow reports the new env var value

hardillb:
Tested using the describe procedure

hardillb (approved):

PR #358, opened 2025-02-28
Do not copy theme into project node_modules

### LINKED ISSUE #356: nr-theme files removed after installing module from palette
### Current Behavior

When installing a module from the palette, files from `@flowfuse/nr-theme` are removed from the filesystem. (However, the theme is still visible in the palette manager under installed nodes)
This causes Node-RED to fall back to the default theme.
Restarting the device-agent service restores the files.

Image

Not a critical problem, just something I've been noticing for some time.

### Expected Behavior

`@flowfuse/nr-theme` should not be removed from filesystem.

### Steps To Reproduce

- Install device agent (e.g., using `raspbian-install-device-agent.sh`)
- Open the Node-RED editor and install a module from palette, e.g.,  `@flowfuse/node-red-dashboad`
- Refresh the browser
- The default Node-RED theme is loaded instead of FlowFuse-Theme
- Error appears in the device logs

### Environment

- FlowFuse version: FlowFuse Cloud
- Node.js version: v18.20.4 & v20.11.1
- npm version: 10.7.0 & 10.9.0
- Platform/OS: Linux RHEL 8.10, Windows Server 2019, Raspberry Pi OS (bookworm)
- Browser: MS Edge, Firefox

### Linked Customers

_No response_

knolleary:
Closes #356 

## Description

The device agent manually copies the theme plugin resources under the `node_modules` directory within the Node-RED working directory.

When installing or removing any other module, `npm` prunes that directory because the theme module isn't listed in the active `package.json` file. Reloading the editor at that point in time leads to a broken theme as the files are missing. Restarting the agent fixes things as the agent recopies the theme over.

At first, I thought the fix would be to ensure the theme is included in the `dependencies` section of `package.json` so that npm wouldn't prune it.

However, on closer inspection, I can't see why the code is copying the theme over at all. The agent already sets `NODE_PATH` to include the agent's own `node_modules` directory where the theme is installed. I have tests with both Node-RED 4.x and 3.1.15 - they both load the theme from that location without any problems.

I can't find a scenario where the theme needs to be copied over (unless, perhaps, it goes back to a NR 2.x issue... but I don't think that's something we need to cater for today).

Hoping @Steve-Mcl may be able to remember back to when this was first added as to why the copy was being done - but also hopeful we can just remove it.

### DIFF
--- a/lib/launcher.js
+++ b/lib/launcher.js
@@ -80,6 +80,11 @@ class Launcher {
                 packageData.dependencies['@flowfuse/nr-assistant'] = `file:${devPath}`
             }
         }
+        if (!packageData.dependencies['@flowfuse/nr-theme']) {
+            // Ensure the theme package is in package.json so that its resources
+            // don't get removed by npm.
+            packageData.dependencies['@flowfuse/nr-theme'] = '*'
+        }
         packageData.version = `0.0.0-${this.snapshot.id}`
         packageData.name = this.snapshot.env.FF_PROJECT_NAME
         if (this.snapshot.name && this.snapshot.description) {
--- a/lib/launcher.js
+++ b/lib/launcher.js
@@ -239,7 +239,13 @@ class Launcher {
                 },
                 projectLink,
                 assistant
-            }
+            },
+            nodesDir: [
+                // This path exists when running the agent as a git clone
+                path.resolve(path.join(__dirname, '..', 'node_modules', '@flowfuse', 'nr-theme')),
+                // This path exists when running the agent as an npm installed package
+                path.resolve(path.join(__dirname, '..', '..', 'nr-theme'))
+            ]
         }
 
         // if licensed, add palette catalogues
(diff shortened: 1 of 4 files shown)


Steve-Mcl:
Since my computer is not "fresh", below are the steps I took to try this out in as clean/isolated way as possible:

1. cloned repo to top level path (avoid any possibility of npm traversal finding theme)
2. switched to branch 356-add-theme-to-package
3. ran `npm i`
4. cleared out `/opt`
5. ran `npm run start -- -o one-time-code -u https://app.flowfuse.com`
6. ran `npm run start`

knolleary:
Okay - I assumed I was missing something here as I couldn't get it to fail in any way that suggested the file copy was needed. But this is running within a dev env setup. Will do a clean standalone test and revisit.

Steve-Mcl:
> I appreciate this won't work for older versions of NR (which don't have all the fixes around `nodesDir` handling you did), but on balance, I think that's okay. What do you think?

Fair compromise to lose the workaround. 
Afterall, NR v4+ is the future and it (NR < v4) does fallback gracefully to the built in theme. 

Will review soon (tomorrow or Thursday likely) unless you need me to jump on it. (let me know if so)

Steve-Mcl (approved):
All good. Tested in clean env.

PR #390, opened 2025-04-29
Import snapshot during otc setup

Steve-Mcl:
### IMPORTANT:
* Do not merge until FF PR [5473](https://github.com/FlowFuse/flowfuse/pull/5473) is approved/merged
* Depends on FF being > 2.15.0 as a way to identify backend supports upload

### TLDR
Adds prompts for importing user flows after device is registered using OTC setup

### Details:
* Adds "@inquirer/input": "^4.1.9" and "@inquirer/select": "^4.2.0"
   * This is from the same set that we currently use for Confirmations (@inquirer/confrim)
* Adds CLI flag `otc-dont-import` to permit automations to skip prompt
* Adds file `lib/cli/flowsImporter.js` that handles the prompting (simplifies integration in `index.js`)
* Refactors logic from `launcher.readPackage` to a utility function `utils.getPackageData` since it will be used by `flowsImporter`
* Adds `loadAndParseJsonFile` and `extractKeyValueFromJsContent` utility functions (full unit tests included)
* Also includes a fix in `test/unit/lib/launcher_spec.js` for flaky test on windows.
* If the device is NOT owned by an application the import step is skipped (as per issue MVP)
* There is a level of sanity checks to avoid uploading duff flows (package exists, flows are an array of objects)
* Flows `credentialSecret` is dug out of the files in the following order:
   * `.config.runtime.json` file
   * If not found, it looks in `settings.json` (for FF device files import)
   * If not found, it scrapes `settings.js`
   * if no `credentialSecret` is found, we set the `credentials` to an empty object and proceed regardless
* Modules are imported from `package.json` but with following caveat:
    * If they are local installed modules (e.g. the module version starts with `file:` instead of a version "^1.0.1" value, it is set to "*"
* The imported snapshot is sent and FFC side sets it as the target (so that upon launching the device, it receives this uploaded snapshot)
    * This is implicit (we don't ask the user, we assume they want to use the imported snapshot)

#### `test/unit/lib/cli/flowsImporter_spec.js`
```
  Flows Importer
    getFlowsFileDetails
      ✔ should identify a valid Node-RED flows file
      ✔ should reject an invalid JSON file
      ✔ should reject a JSON file that is not a Node-RED flows file
    getFlowFiles
      ✔ should find Node-RED flows files in a directory
      ✔ should ignore files that are not flows files
      ✔ should return empty array for invalid directory
      ✔ should return empty array if package.json does not exist
    getDirDetails
      ✔ should return valid directory details
      ✔ should return invalid details for directory without flows
    askSelectFlowsFile
      ✔ should prompt user to select a flows file
      ✔ should return false if user selects skip
      ✔ should return false if no flow files are provided
      ✔ should prompt user to select a flows file
    askImportDirectory
      ✔ should return directory details if user selects a suggested directory
      ✔ should prompt for custom path if user selects custom option
      ✔ should prompt with confirm dialog if no suggested dirs are provided
      ✔ should return skip result if user cancels custom path entry
    flowImport
      ✔ should return null if user skips import
      ✔ should return null if user does not select a flows file
      ✔ should import flows, credentials and package data
      ✔ should get credential secret from settings.json
      ✔ should get credential secret from settings.js
      ✔ should handle missing credentials file
```

#### `test/unit/lib/util_spec.js`
```
  utils
    getPackageData
      ✔ should correctly parse package data
    extractKeyValueFromJsContent
      ✔ should extract the correct value for a given key
      - should extract the correct value for a given key in a minified settings file
      ✔ should return null if the key is not found
      ✔ should return null if the key is // commented out
      ✔ should return null if the key is /* commented */ out
    loadAndParseJsonFile
      ✔ should correctly parse a valid JSON file
      ✔ should return null if the file does not exist
      ✔ should return null if the file contains invalid JSON
```

## Related Issue(s)

Owner: #387 
Parent:  https://github.com/FlowFuse/flowfuse/issues/5366

## Checklist

    - [ ] Upgrade instructions
    - [x] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,10 +258,10 @@ Web UI Options
 
 Setup command
 
-  -o, --otc string         Setup device using a one time code
-  -s, --otc-dont-start     Do not start the agent after setup
-  -n, --otc-dont-import    Do not ask to import Node-RED flows during setup
-  -u, --ff-url url         URL of FlowFuse. Required for setup
+  -o, --otc string    Setup device using a one time code
+  --otc-no-start      Do not start the agent after setup
+  --otc-no-import     Do not ask to import Node-RED flows during setup
+  -u, --ff-url url    URL of FlowFuse. Required for setup
 
 Global Options
 
--- a/README.md
+++ b/README.md
@@ -258,8 +258,10 @@ Web UI Options
 
 Setup command
 
-  -o, --otc string   Setup device using a one time code
-  -u, --ff-url url   URL of FlowFuse. Required for setup
+  -o, --otc string         Setup device using a one time code
+  -s, --otc-dont-start     Do not start the agent after setup
+  -n, --otc-dont-import    Do not ask to import Node-RED flows during setup
+  -u, --ff-url url         URL of FlowFuse. Required for setup
 
 Global Options
 
--- a/index.js
+++ b/index.js
@@ -133,9 +133,9 @@ Please ensure the parent directory is writable, or set a different path with -d`
             print('Success! This Device can be launched at any time using the following command:', figures.tick)
             print(runCommandInfo.join(' '), ' ')
             if (!options.otcNoImport) {
-                // Support for importing flows during initial state check-in was added after 2.15.0.
+                // Support for importing flows during initial state check-in was added after 2.16.0.
                 const ffVersion = deviceSettings.meta?.ffVersion?.replace(/[^0-9.]/g, '') || '0.0.0' // Strip suffixes like -beta.1
-                const ffSupportsImport = (ffVersion && semver.gt(ffVersion, '2.15.0'))
+                const ffSupportsImport = (ffVersion && semver.gt(ffVersion, '2.16.0'))
 
                 if (ffSupportsImport) {
                     const home = process.env.HOME || process.env.USERPROFILE || process.env.HOMEPATH || '/'
--- a/index.js
+++ b/index.js
@@ -15,7 +15,10 @@ const { AgentManager } = require('./lib/AgentManager')
 const { WebServer } = require('./frontend/server')
 const ConfigLoader = require('./lib/config')
 const webServer = new WebServer()
+const figures = require('@inquirer/figures').default
 const confirm = require('@inquirer/confirm').default
+const print = (message, /** @type {figures} */ figure = figures.info) => console.info(figure ?? figures.info, message)
+const flowImport = require('./lib/cli/flowsImporter').flowImport
 
 function main (testOptions) {
     const pkg = require('./package.json')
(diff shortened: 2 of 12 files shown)


Steve-Mcl:
## Acceptance Windows

- [x] can skip at dir choice level
- [x] suggests exiting `~/.node-red` dir
    - [x] lists flows from dir
    - [x] imports flows
    - [x] can skip at flow choice level
- [x] suggests exiting `/opt/flowfuse-device/project` dir
    - [x] lists flows from dir
    - [x] imports flows
        - [x] cleans old flows in /opt/flowfuse-device after import
    - [x] can skip at flow choice level
    
 Performed by @Steve-Mcl

Steve-Mcl:
## Acceptance Linux

- [ ] can skip at dir choice level
- [ ] suggests exiting `~/.node-red` dir
    - [ ] lists flows from dir
    - [ ] imports flows
    - [ ] can skip at flow choice level
- [ ] suggests exiting `/opt/flowfuse-device/project` dir
    - [ ] lists flows from dir
    - [ ] imports flows
        - [ ] cleans old flows in /opt/flowfuse-device after import
    - [ ] can skip at flow choice level

Steve-Mcl:
## Acceptance MacOS

- [ ] can skip at dir choice level
- [ ] suggests exiting `~/.node-red` dir
    - [ ] lists flows from dir
    - [ ] imports flows
    - [ ] can skip at flow choice level
- [ ] suggests exiting `/opt/flowfuse-device/project` dir
    - [ ] lists flows from dir
    - [ ] imports flows
        - [ ] cleans old flows in /opt/flowfuse-device after import
    - [ ] can skip at flow choice level

README.md
```diff
@@ -258,8 +258,10 @@ Web UI Options
 
 Setup command
 
-  -o, --otc string   Setup device using a one time code
-  -u, --ff-url url   URL of FlowFuse. Required for setup
+  -o, --otc string         Setup device using a one time code
+  -s, --otc-dont-start     Do not start the agent after setup
+  -n, --otc-dont-import    Do not ask to import Node-RED flows during setup
```
knolleary, line 263: I had proposed in the FF pr to change these flags - without realising only one of them is 'new' in this PR, which makes it a bit trickier to propose changes here.

We haven't released the `-s` flag yet, so we can still change if needed.

My concern is the short versions `-s` and `-n` feel a little random - would we be better to not have short-codes and just have the long form for these two?

knolleary (commented):
Looks okay - open question on the specific cli args being added.

README.md
Steve-Mcl: sensible. Will address.

Steve-Mcl:
@knolleary CLI args updated - ready for final review.

@ppawlowski this may be of interest for you.  We have updated the no start and no import  CLI flags to more familiar format.

```log
Setup command

  -o, --otc string    Setup device using a one time code
  --otc-no-start      Do not start the agent after setup
  --otc-no-import     Do not ask to import Node-RED flows during setup
  -u, --ff-url url    URL of FlowFuse. Required for setup
  ```

File: index.js [hunk 2]
```diff
@@ -111,28 +114,89 @@ Please ensure the parent directory is writable, or set a different path with -d`
...
+            const runCommandInfo = ['flowfuse-device-agent']
+            if (options.dir !== '/opt/flowfuse-device') {
+                runCommandInfo.push(`-d ${options.dir}`)
+            }
+            print('Success! This Device can be launched at any time using the following command:', figures.tick)
+            print(runCommandInfo.join(' '), ' ')
+            if (!options.otcNoImport) {
+                // Support for importing flows during initial state check-in was added after 2.15.0.
+                const ffVersion = deviceSettings.meta?.ffVersion?.replace(/[^0-9.]/g, '') || '0.0.0' // Strip suffixes like -beta.1
+                const ffSupportsImport = (ffVersion && semver.gt(ffVersion, '2.15.0'))
```
Steve-Mcl, line 138:
```suggestion
                // Support for importing flows during initial state check-in was added after 2.16.0.
                const ffVersion = deviceSettings.meta?.ffVersion?.replace(/[^0-9.]/g, '') || '0.0.0' // Strip suffixes like -beta.1
                const ffSupportsImport = (ffVersion && semver.gt(ffVersion, '2.16.0'))
```

Current version of FF is 2.16 - this should check for > 2.16

knolleary (approved):

PR #402, opened 2025-05-15
Only use local HOME env var if not passed from settings

hardillb:
## Description

Customer running in OpenShift with no HOME env var set in the container.

They tried to set it in the FlowFuse device settings, but this code overwrites with local even if not set.

This only uses the local version if

1. it's present
2. it's not come from settings

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/lib/launcher.js
+++ b/lib/launcher.js
@@ -475,9 +475,13 @@ class Launcher {
 
         // should set HOME env var
         if (process.platform === 'win32') {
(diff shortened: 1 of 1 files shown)


Steve-Mcl (approved):
Makes sense.

PR #426, opened 2025-06-25
Add blueprint library plugin and rename team library plugin

knolleary:
## Description

Adds the blueprint library plugin. This is the same plugin as added to the launcher via https://github.com/FlowFuse/nr-launcher/pull/345 - with one functional addition to support proxies.

I also renamed the existing Team Library plugin - all internal details, but no reason to keep it under the old name.


lib/launcher.js
```diff
@@ -274,6 +274,23 @@ class Launcher {
...
+                label: 'Blueprints',
+                icon: 'font-awesome/fa-map-o',
+                types: ['flows'],
+                readOnly: true,
+                forgeURL: this.config.forgeURL,
+                teamID: settings.flowforge.teamID,
+                token: this.config.token
+            })
+        }
+        console.log(settings.nodesDir)
```
hardillb, line 293:
```suggestion
```

PR #565, opened 2026-01-28
feat!: Run Device Agent in container as unpriviledged user

### LINKED ISSUE #562: Docker container should run on non-root user
### Description

The goal of this task is to validate the possibility running the Device Agent in a container as an unprivileged user.

### Requested By

https://app-eu1.hubspot.com/contacts/26586079/record/0-2/210431677688

ppawlowski:
## Description

BREAKING CHANGE
This pull request changes the way how the device-agent runs inside a container. Instead of using a root user, a custom `flowfuse` one is created during the container build process.

Additionally, it:
* adds a workflow that ensures the container build process is validated against every PR that introduces changes to Docker-related files
* runs Static Application Security Test scan against the built container and pushes results to the GitHub Advanced Security
* Adjusts the node package build pipeline to skip execution if Docker-related files are modified

## Related Issue(s)

Closes https://github.com/FlowFuse/device-agent/issues/562

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


hardillb (commented):
My problem with this is it is potentially a breaking change, so we should merge and apply it the same time we change the default NodeJS version (to 20) at the next major release (4.0.0?).

I'm pretty sure this will break users that are extending the flowfuse/device-agent container (e.g. "windows") as the active user will change from `root` to a user that can not install `apk` packages or create new directories.

There is a workaround for the existing container which is to pass `-u 1000:1000` to set the uid and gid to 1000 so as not to run as root.

Need to think about this some more

knolleary:
Given this is a breaking change, what exactly does a user need to do to upgrade?
If this needs to be a major version bump, what other issues need to be completed before we do that release?

ppawlowski:
I've added an entry to the README regarding changes required when a host directory is used for data persistence.

hardillb (approved):

knolleary:
@ppawlowski can you check the merge conflict on this PR?

ppawlowski:
> @ppawlowski can you check the merge conflict on this PR?

fixed.

PR #664, opened 2026-06-24
docs: Revise README for Device Agent v4 changes

knolleary:
Add notes on breaking changes at start of readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ sudo npm install -g @flowfuse/device-agent
 npm install -g @flowfuse/device-agent
 ```
 
-#### Docker
+### Docker
 
 We publish a Docker container for the Device Agent as `flowfuse/device-agent` on DockerHub.
 
--- a/README.md
+++ b/README.md
@@ -3,12 +3,20 @@
 This module provides an agent that runs Node-RED instances deployed from the
 FlowFuse platform.
 
+> [!IMPORTANT]
+> The v4 release of the device agent brings some potentially breaking changes:
+>
+> - The `latest` docker tag now points at a NodeJS 22 based container
+> - The docker container runs as a non-root user. If you mount an external host directory, ensure it is accessible to the user.
+>
+> More information about these changes are available in the [docker section](#docker).
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,428 +1,439 @@
...
+ - NodeJS v18 or later. NodeJS v22 is recommended as the minimum version to use as that is what Node-RED v5 requires.
+ - A FlowFuse platform to connect to
+
+For users that require NodeJS v14 and v16 support the v2.x stream will work.
+
+## Supported Operating Systems
+
+The Device Agent can be installed on most Linux distributions, Windows, and MacOS.
+
+## Installing the Device Agent
```
hardillb, line 25: We don't appear to mention the native installer at all in here?

hardillb (approved):
Looks good, only question is that we don't mention the native installer at all in the README that I can spot

README.md
knolleary: Good point - working on it

ppawlowski (approved):

README.md
knolleary: Link to installer docs added

README.md
```diff
@@ -6,43 +6,51 @@ FlowFuse platform.
...
 
-### Windows
+#### Windows
 
 ```bash
 npm install -g @flowfuse/device-agent
 ```
 
-### Docker
+#### Docker
```
ppawlowski: This section is now a sub-paragraph of the `Installing with npm` one. IMO they should be on the same level.

knolleary: Yes - you are right.

README.md
```diff
@@ -1,428 +1,447 @@
...
+sudo npm install -g @flowfuse/device-agent
+```
+
+#### Windows
+
+```bash
+npm install -g @flowfuse/device-agent
+```
+
+#### Docker
```
knolleary:
```suggestion
### Docker
```

knolleary:
GitHub found a mix of line endings in the file and has standardized them for me... hence the rather large looking diff.

PR #668, opened 2026-06-25
feat: Add insights support

### LINKED ISSUE #625: Support MCP Features and MCP Tool calling for Expert Insights
### Description

To realise Expert Insights on a remote instance, support must be present over the existing MQTT connection.

The proposal is to extend the device-agent's MQTT command channel with three new commands that relay MCP calls to MCP servers hosted inside the device's Node-RED instance:

* `mcp:get-features` - discover capabilities, tools, resources, resource templates and prompts across one or more MCP endpoints.
* `mcp:call-tool` - invoke a named tool on a given endpoint with arguments.
* `mcp:read-resource` - read a named resource from a given endpoint.

I propose a new MCP helper class (packages/device-agent/lib/mcp.js) wraps the functionality of the @modelcontextprotocol/sdk Client and StreamableHTTPClientTransport such that the MQTT handler simply calls `this.mcp.getFeatures(msg.payload?.mcpEndPoints)`

Additionally, to minimise timeouts, I also propose device agent also supports a new command `get-liveState` - this provides parity with the instance models `liveState()` function and will prevent unnecessary calls (and subsequent  timeouts) when the mcp:get-features request is made.

NOTE: Coupled with the centrally bridged MQTT Broker for Expert, this will provide a path enabling Insights on self hosted devices too.

### Epic/Story

_No response_

Steve-Mcl:
### **TLDR**

Add support for MCP tasks (get features, call tool, read resource) command-able via the established command-response pattern (over MQTT) - thus permitting AI Insights Agent to request/execute MCP things via bridge.

Add support for requesting live-state from the device (via command-response pattern (over MQTT)) so that live state can be queried before attempting (and timing out) to get MCP features / execute MCP thing.

NOTE: To avoid adding the MCP SDK to the dependencies, they are loaded from the `node_modules` directory of the `nr-mcp-server-nodes`. The logic/reasoning is :-
1. the MCP SDK is a fast changing package
2. if mcp server nodes are not installed, there is no need for this lib to be installed or instantiated
3. version parity with MCP server nodes is maintained

### Details
**MCP SDK Integration and Lifecycle Management:**

* The MQTT client (`lib/mqtt.js`) now instantiates an `MCP` helper, making MCP features available for MQTT command handling. [[1]](diffhunk://#diff-22d60ae947b2cc2a1180672627a138f507749c3f98792237cd599d075ec38998R1-R6) [[2]](diffhunk://#diff-22d60ae947b2cc2a1180672627a138f507749c3f98792237cd599d075ec38998R52-R59)
* The `Launcher` class (`lib/launcher.js`) now resets any cached MCP SDK reference both when starting and stopping, ensuring that the SDK is always up-to-date and not reused if the underlying files have changed. [[1]](diffhunk://#diff-a4ef8d754c46f33ab02a620e91925ac615d7412f2f80edf06b2425cba5ff3104R552-R557) [[2]](diffhunk://#diff-a4ef8d754c46f33ab02a620e91925ac615d7412f2f80edf06b2425cba5ff3104R748-R752)

**New MQTT Command Support:**

* The MQTT client now handles several new commands: `get-liveState`, `mcp:get-features`, `mcp:call-tool`, and `mcp:read-resource`, delegating these to the MCP helper and sending structured responses or errors as appropriate.

**MQTT Command Response Improvements:**

* The `sendCommandResponse` method now includes `userProperties` from the request in the response message, allowing for richer metadata propagation. [[1]](diffhunk://#diff-22d60ae947b2cc2a1180672627a138f507749c3f98792237cd599d075ec38998R388) [[2]](diffhunk://#diff-22d60ae947b2cc2a1180672627a138f507749c3f98792237cd599d075ec38998R400)

**Dependency Management:**

* The `path` module is now required in `lib/mqtt.js` to support MCP instantiation.

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


lib/mqtt.js
```diff
@@ -330,6 +385,7 @@ class MQTTClient {
     sendCommandResponse (request, response) {
         const correlationData = request?.correlationData
         const responseTopic = request?.responseTopic || this.responseTopic
+        const userProperties = request?.userProperties || {}
```
hardillb, line 388: Confused with this name that it was `msg.properties.userProperties` (MQTTv5) not just a filed on the JSON in the `msg.payload`

lib/mqtt.js
```diff
@@ -144,6 +154,51 @@ class MQTTClient {
                     await this.handleActionRequest(msg)
                 } else if (msg.command === 'reportPackages') {
                     this.reportPackages = true
+                } else if (msg.command === 'get-liveState') {
```
hardillb, line 157: should this have a `mcp:` prefix?


```suggestion
                } else if (msg.command === 'mcp:get-liveState') {
```

hardillb (commented):
Just a question about a command to clear up

And confirmation that this shouldn't need MQTTv5 for the userProperties stuff

lib/mqtt.js
Steve-Mcl: yes, you are correct. As device agent is currently (from inception) using the MQTT default protocol version v4 (3.1) v5 properties are not available. What we do to achieve req-resp pattern is to nest the payload in the MQTT payload (`request` object) and add the correlationData for tracking responses.
This addition here includes userProperties so that we have the ability to pass through (and pass back) any user properties that need to be carried through,

lib/mqtt.js
Steve-Mcl: no, but i can totally see why you might think that.

This is more general purpose - akin to the **nr-launcher**s `instance.liveState` method.
This is essentially the equivalent via MQTT.

For insights, it is called before ever attempting to run MCP get features to avoid unnecessary timeouts when the device is shutdown or not running.

See the instance and device guards in this section of code in the partner PR:

https://github.com/FlowFuse/flowfuse/pull/7604/changes#diff-9d00faf29e4cc28ba661a77876079c18c256a7b24a48ce99f73715447a2a557cR434-R458

Steve-Mcl:
@hardillb hopefully answered yer questions (re-requested review)

hardillb (approved):

hardillb:
@Steve-Mcl approved, but @cstns should look at the `mcp.js` file

lib/mcp.js
```diff
@@ -0,0 +1,466 @@
...
+ */
+function endpointToUrl (endpoint, host, port) {
+    try {
+        const u = new URL(endpoint)
+        if (u.protocol === 'http:' || u.protocol === 'https:') {
+            return u
+        }
+    } catch (_) { /* fall through to local-path handling */ }
+    const pathPart = endpoint.startsWith('/') ? endpoint : `/${endpoint}`
+    return new URL(`http://${host || '127.0.0.1'}:${port}${pathPart}`)
```
cstns: having `:` always present might blow up if port is null/undefined ie k8/docker. a quick fix would be to coerce port to an empty string?

lib/mcp.js
```diff
@@ -0,0 +1,466 @@
...
+ * @returns {McpEndpoint}
+ *   `key` is what the result map is keyed by; `skip` is set when the endpoint
+ *   should not be contacted (e.g. unsupported auth scheme).
+ */
+function normalizeEndpoint (spec, host, port) {
+    if (typeof spec === 'string') {
+        return {
+            key: spec,
+            url: endpointToUrl(spec, host, port).toString(),
+            headers: { 'Content-Type': 'application/json' },
```
cstns, line 128: shouldn't we be including the Accept header as well?

cstns: also, shouldn't we be adding the auth header for this return case?

lib/mcp.js
```diff
@@ -0,0 +1,466 @@
...
+    /**
+     * Connect a fresh MCP client to the given (already-normalised) endpoint spec.
+     * Caller is responsible for `client.close()` when finished.
+     * @param {McpEndpoint} endpoint
+     * @returns {Promise<McpClient>} a connected MCP Client instance
+     */
+    async _connect (endpoint) {
+        const { Client, StreamableHTTPClientTransport } = this._loadSdk()
+        const transport = new StreamableHTTPClientTransport(endpoint.url, {
+            requestInit: { headers: endpoint.headers }
```
cstns, line 270: shouldn't we add `sessionIdGenerator: undefined`? all incoming requests will be stateless if i'm not mistaken so there's no reason to add additional overhead for the server to track sessions

lib/mcp.js
Steve-Mcl: On k8s / docker, there is still a node-red running on http ip:port

@hardillb correct me if I am wrong?

lib/mcp.js
Steve-Mcl: `sessionIdGenerator` is an option of the server transport. This code is the client transport (there is no sessionIdGenerator, adding it would be silently ignored.)

lib/mcp.js
Steve-Mcl: Perhaps a compromise option would - default to 1880?

e.g. 

```suggestion
    return new URL(`http://${host || '127.0.0.1'}:${port || 1880}${pathPart}`)
```

Maybe a default in the contructoir too?
```
constructor ({ nodeRedUserDir, host = '127.0.0.1', port = 1880 } = {}) {
```

WDYT @cstns @hardillb

hardillb: Yes, Node-RED should always be running on port 1880 inside the container for both docker and k8s

lib/mcp.js
Steve-Mcl: This is what we have in mcp server nodes so I know its not an issue but I take your point.

The thing is, a bit of digging reveals:

https://unpkg.com/@modelcontextprotocol/sdk/dist/esm/client/streamableHttp.js - look in the async send(...) method.

`accept` and `content-type` are force set by the SDK via `.set()` on every POST. So adding Accept is a noop, and the Content-Typesset here is kinda redundant! Who knew.

If you make a suggestion, I will add it (for completeness) _(It is a no harm thang either way.)_

> also, shouldn't we be adding the auth header for this return case?

No, just skip auth if not provided - i.e. a URL "string" only would mean no auth is offered (hence the early return)

The real auth is applied to headers (when a token is provided) a few lines down

Steve-Mcl:
@cstns check answers to comments & reply/resolve/suggest please - wanna get this moving.

lib/mcp.js
cstns: My bag, got them mixed up 🤦‍♂️

cstns (approved):

PR #687, opened 2026-07-10
fix: Merge catalogue and npmrc for hosted instance bound devices

### LINKED ISSUE #677: Hosted Instance Bound Devices not getting Certified Nodes Catalogues
### Current Behavior

If a Remote Instance is bound to a Hosted Instance (rather than to an Application) take their catalogue and `.npmrc` file data from the snapshot not the devices settings.

### Expected Behavior

Not sure, this was explicitly chosen when it was added.

I do question if a Instance bound device should be editable directly, the point it that it is supposed to follow it's bound Hosted Instance.

### Steps To Reproduce

Create device, bind to hosted instance
Create snapshot from Hosted Instnce
Put Device into Developer mode and open editor

### Environment

- FlowFuse version: 2.32.4
- Node.js version: 24
- npm version: na
- Platform/OS: all
- Browser: na

### Linked Customers

hardillb:
fixes #677

## Description

If this is a hosted instance bound device then it merges the settings and the snapshot npmrc and catalogue lists.

This means devices shold see certified nodes if enabled

We merge both and remove duplicate entries.

Will need a device agent release to apply.

## Related Issue(s)

#677

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/lib/launcher.js
+++ b/lib/launcher.js
@@ -349,7 +349,7 @@ class Launcher {
                     settings.editorTheme.palette.catalogues = this.snapshot.settings.palette.catalogue
                     // need to merge in settings catalogues and remove duplicates
                     const temp = [...(this.settings?.palette?.catalogues || []), ...(this.snapshot.settings?.palette?.catalogue || [])]
-                    settings.editorTheme.palette.catalogue = temp.filter((item, pos, self) => {
+                    settings.editorTheme.palette.catalogues = temp.filter((item, pos, self) => {
                         return self.indexOf(item) === pos
                     })
                 }
--- a/lib/launcher.js
+++ b/lib/launcher.js
@@ -348,7 +348,7 @@ class Launcher {
                 if (this.snapshot?.settings?.palette?.catalogue !== undefined) {
                     settings.editorTheme.palette.catalogues = this.snapshot.settings.palette.catalogue
                     // need to merge in settings catalogues and remove duplicates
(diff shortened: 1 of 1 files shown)


knolleary:
Some test failures to attend to

hardillb:
Tests passing now

lib/launcher.js
```diff
@@ -347,6 +347,11 @@ class Launcher {
             if (this.project) {
                 if (this.snapshot?.settings?.palette?.catalogue !== undefined) {
                     settings.editorTheme.palette.catalogues = this.snapshot.settings.palette.catalogue
+                    // need to merge in settings catalogues and remove duplicates
+                    const temp = [...(this.settings?.palette?.catalogues || []), ...(this.snapshot.settings?.palette?.catalogue || [])]
+                    settings.editorTheme.palette.catalogue = temp.filter((item, pos, self) => {
```
knolleary: Given the code on line 349, should this be `catalogues`? We have really made life difficult for ourselves the day one of us made that typo and it didn't get spotted until it was too late.

```suggestion
                    settings.editorTheme.palette.catalogues = temp.filter((item, pos, self) => {
```

knolleary (changes requested):
Added a comment early, but forgot to do it as a review.

Concerned theres a catalogue/catalogues mismatch here - want to be very certain we get this right first time.

lib/launcher.js
hardillb: yeah looks like I missed one

knolleary (approved):

PR #702, opened 2026-07-23
feat(installer): Add possibility to provide custom CA bundle certificate

### LINKED ISSUE #698: Installer: Add possibility to onboard a Remote Instance to the sel-hosted platform running with TLS issued by a private CA
### Description

The goal of this task is to add the possibility of onboarding the Device Agent using the Installer when the FlowFuse Platform is running with TLS certificates issued by a private Certificate Authority.
Two aspects should be considered:
1. The setup phase will fail with the ""Unable to connect to FlowFuse Platform" error
2. Once installed, the agent runtime must also properly connect to the platform

### Requested By

From the support thread: https://app-eu1.hubspot.com/help-desk/26586079/view/233410279/ticket/424030742746/thread/20613191903#email

ppawlowski:
## Description

This pull request introduces the `--ca-cert` flag to the Device Agent Installer that allows to pass the Certificate Authority certificate bundle file to be trusted by the Device Agent.
This change allows to quickly onboard the Remote Instance using the Installer in environments where self-signed certificates have been used to deploy the FlowFuse Platform.

## Related Issue(s)

Closes https://github.com/FlowFuse/device-agent/issues/698

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/installer/README.md
+++ b/installer/README.md
@@ -44,6 +44,7 @@ powershell -c Unblock-File -Path .\flowfuse-device-agent-installer.exe
 | `--dir` | `-d` | `/opt/flowfuse-device` (Linux/macOS) or `C:\opt\flowfuse-device` (Windows) | Installation directory for the device agent |
 | `--port` | `-p` | `1880` | TCP port for the device agent (1025–65535). Service name is suffixed with the port, e.g., `flowfuse-device-agent-1880`. |
 | `--uninstall` | | `false` | Uninstall the device agent |
+| `--ca-cert` | | *optional* | Path to a CA certificate bundle (PEM) the Device Agent should trust |
 | `--update-nodejs` | | `false` | Update bundled Node.js to specified version |
 | `--update-agent` | | `false` | Update the Device Agent package to specified version |
 | `--debug` | | `false` | Enable debug logging |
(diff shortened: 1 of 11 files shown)


allthedoll (approved):
This looks good to merge to me as is. The flag is off by default so ship risk is low, and reusing the stored cert on reinstall is ✨.

I think you should add a smol unit test for `InstallCACertificate`, though, which can be done in another iteration. Something for the empty/missing/not-a-file cases so a future change can't silently skip the CA.

I have two other points to consider:
- `--ca-cert` is ignored on `--update-agent` where it'd be nice to have a doc line or comment
- On Linux/MacOS the file is read into memory and _also_ copied via temp file, so IMHO the `os.ReadFile` could probably be dropped since the temp file is what gets copied?

This is really nice work though, @ppawlowski and I appreciate the code comments.

ppawlowski:
> --ca-cert is ignored on --update-agent where it'd be nice to have a doc line or comment

 Added a comment to readme regarding when the `--ca-cert` file is used.

> On Linux/MacOS the file is read into memory and also copied via temp file, so IMHO the `os.ReadFile` could probably be dropped since the temp file is what gets copied?

While that's true, I use this approach in a few other places too. The goal is to not pass an operator-provided path straight to `cp`. The `os.ReadFile` function handles any filename fine when reading, but the `exec.Command("sudo","cp", srcPath, …)` passes that path as an argument to `cp`, and a source path such as `-rf` or `--something-nasty` would be parsed by `cp` as an option rather than a file, plus relative paths depend on the working directory. Reading the file and re-creating it under a predictable temp path/name guarantees the copy step gets a known-safe argument and won't fail on a weird `--ca-cert` value.
Additionally, for small files like certificates, the extra read/write is irrelevant to performance.

PR #708, opened 2026-07-29
chore(installer): Update installer for async registration flow

knolleary:
Update the installer to work with the new async registration flow.

There is a big chicken/egg issue when it comes to testing this PR. It installs latest device agent code from npm - that's no good when you want to test installer changes alongside unpublished device agent changes.

I found a hacky workaround by temporarily modifying the installer code to install from a hardcoded local path rather than npm. Not entirely satisfactory.

But, it is true to say these installer updates do not need to merge at the same time as the setup PR. We can do a release of the updated device agent, then finish verification of the installer against the newly published version.

The main changes are:

 1. Adds some utils for formatting console output (similar to chalk in the node.js code). This allows for more consistent output. But - see the note below about future work.
2. The installer no longer requires the OTC being provided - as the new device agent can handle that for you.
3. It disables the ability to paste in a full device.yml file - again, see below.

### Future work

The original intent was to rework the whole sequence of steps the installer takes - to give the user more control. I have a local branch with that 90% done, but its a lot of work to fully test. So I've reduced the scope a lot to keep the same level of opinionated behaviour, but with the async registration supported.

I will raise separate issues to identify what needs doing next, but that doesn't block this iteration.


knolleary:
Given the installer release doesn't use release-please currently, I think I should make this a `chore:` PR to avoid bumping the Device Agent version unnecessarily.

Steve-Mcl:
> Given the installer release doesn't use release-please currently, I think I should make this a `chore:` PR to avoid bumping the Device Agent version unnecessarily.

Agree

Steve-Mcl (approved):

PR #726, opened 2026-08-21
feat(installer): Ask for installation directory in the interactive mode

### LINKED ISSUE #725: Ask user about installation directory in the interactive mode
### Description

When installing the Device Agent with the Installer in interactive mode, the installer should ask the user about the installation path (location on the local storage where Device Agent files should be installed).
This should not brake existing installation path when the directory is passed to the Installer with the `--dir` flag.

### Requested By

ppawlowski:
## Description

This pull request extends the Device Agent Installer interactive mode by asking a user about installation directory suggesting the default install directory.
To maintain the unattended installation offered by the FlowFuse Platform, the interactive mode is considered when not OTC value is provided.

## Related Issue(s)

Closes https://github.com/FlowFuse/device-agent/issues/725

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


Steve-Mcl (approved):
ok, so while this works well (tested several times) what is now a happy side-effect of offering which directory to install to promotes multiple installations - which is nice - however we dont ask the user for a port number to run it on (relying on the --port arg)

I would strongly encourage us to also ask for a port number FIRST then offer a suitable dir like `opt/flowfuse-device-1889` as a default as a tie up with PR

However, I will not block this on that ^ so this is approved from my POV.

PR #728, opened 2026-08-25
feat(installer): Prompt user if system service should be installed in the interactive mode

### LINKED ISSUE #727: Ask user if Device Agent should be started on boot as a system service
### Description

The goal of this task is to give a user possibility to decide wheather the Device Agent should start on system boot as s system service.

### Requested By

ppawlowski:
## Description

This pull request introduces a user prompt for system service installation when installer runs in the interactive mode (OTC is not provided).
Additionally, it updates the update and uninstall scenarios, when a Device Agent process can run in the background, not started as a system service.

## Related Issue(s)

Closes https://github.com/FlowFuse/device-agent/issues/727

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


installer/go/pkg/utils/utils.go
```diff
@@ -1203,6 +1344,47 @@ func ShowInstallSummary(installMode, url, workDir string) {
...
+	logger.Info("")
+
+	switch runtime.GOOS {
+	case "windows":
+		if caCertPath != "" {
+			logger.Info(`  set NODE_EXTRA_CA_CERTS=%s`, caCertPath)
+		}
+		logger.Info(`  set PATH=%s;%%PATH%%`, nodeBinDir)
+		logger.Info(`  %s --dir "%s" --port %d`,
+			filepath.Join(nodeBinDir, "flowfuse-device-agent.cmd"), workDir, port)
```
Steve-Mcl: This might confuse those who favour powershell



I think a reasonable compromise would be to print both for windows

```suggestion
		agentCmd := filepath.Join(nodeBinDir, "flowfuse-device-agent.cmd")
		logger.Info("  Command Prompt (cmd):")
		if caCertPath != "" {
			logger.Info(`    set NODE_EXTRA_CA_CERTS=%s`, caCertPath)
		}
		logger.Info(`    set PATH=%s;%%PATH%%`, nodeBinDir)
		logger.Info(`    %s --dir "%s" --port %d`, agentCmd, workDir, port)
		logger.Info("")
		logger.Info("  PowerShell:")
		if caCertPath != "" {
			logger.Info(`    $env:NODE_EXTRA_CA_CERTS = "%s"`, caCertPath)
		}
		logger.Info(`    $env:Path = "%s;" + $env:Path`, nodeBinDir)
		logger.Info(`    & "%s" --dir "%s" --port %d`, agentCmd, workDir, port)
```

This would output:

Steve-Mcl: Additionally, I would probably recommend that the user is instructed to use a fresh non-root terminal since - something like "Run these in a new, non-Administrator terminal - in most cases, the agent should not need to run elevated."

installer/go/pkg/utils/utils.go
```diff
@@ -1359,12 +1359,20 @@ func ShowManualStartInstructions(nodeBinDir, workDir, caCertPath string, port in
...
-			filepath.Join(nodeBinDir, "flowfuse-device-agent.cmd"), workDir, port)
+		logger.Info(`    set PATH=%s;%%PATH%%`, nodeBinDir)
+		logger.Info(`    %s --dir "%s" --port %d`, agentCmd, workDir, port)
+		logger.Info("")
+		logger.Info("  PowerShell:")
+		if caCertPath != "" {
+			logger.Info(`    $env:NODE_EXTRA_CA_CERTS = "%s"`, caCertPath)
+		}
+		logger.Info(`    $env:Path = "%s;" + $env:Path`, nodeBinDir)
+		logger.Info(`    & "%s" --dir "%s" --port %d`, agentCmd, workDir, port)
```
Steve-Mcl, line 1375: Did you see my [other comment](https://github.com/FlowFuse/device-agent/pull/728#discussion_r3855218948):

> Additionally, I would probably recommend that the user is instructed to use a fresh non-root terminal since - something like "Run these in a new, non-Administrator terminal - in most cases, the agent should not need to run elevated."

I think this would be good to try and encourage users not to fire up the agent as root/admin
WDYT?

ppawlowski: Are we 110% sure that the device agent will run properly on a non-privileged account?

Steve-Mcl: Nothing is 110% but we should not (however inadvertently) ask or encourage users to "run this command" within and elevated command/powershell

FWIW, I did (the outputted commands in PS/CMD ) test as a regular user - it works (but then I am a dev so i don't really count due to years of system hacks!)

ppawlowski: That's the reason I put a simple message, `run in the command line`, without specifying whether this should be executed by a user with elevated privileges. Not a fan of saying explicitly that it is safe to run this as an unprivileged user, since we, as code owners, cannot guarantee that this will work.

installer/go/pkg/utils/utils.go
```diff
@@ -1203,6 +1344,55 @@ func ShowInstallSummary(installMode, url, workDir string) {
...
+// the foreground, for installations where no system service was created.
+//
+// Parameters:
+//   - nodeBinDir: the directory holding the bundled node and agent executables
+//   - workDir: the installation directory, passed to the agent as --dir
+//   - caCertPath: the installed CA bundle, or "" when none was provided
+//   - port: the TCP port the agent should listen on
+func ShowManualStartInstructions(nodeBinDir, workDir, caCertPath string, port int) {
+	logger.Info("")
+	logger.Info("To start the FlowFuse Device Agent manually, run in the command line:")
```
Steve-Mcl:
```suggestion
	logger.Info("To start the FlowFuse Device Agent manually, run the command in a new terminal:")
```

How about this instead?

Steve-Mcl: this way, we leave the user/elevated choice to user?

Steve-Mcl (approved):

PR #739, opened 2026-09-03
feat(installer): Prompt for the Device Agent port in interactive mode

### LINKED ISSUE #730: Installer: Prompt for a port number in the interactive mode
### Description

As a Remote Instance integrator using the Device Agent Installer in interactive mode, I want to be able to define a custom TCP port to which the Device Agent will bind.

### Requested By

https://github.com/FlowFuse/device-agent/pull/726#pullrequestreview-5029157999

ppawlowski:
## Description

This pull request adds a port question to the installer's interactive mode, so a Remote Instance integrator can choose which TCP port the Device Agent binds to without having to know the `--port` flag up front. The installer asks for the
port straight after the permission check and re-asks until the answer is a free port in the 1025-65535 range. Passing `--port` still works unchanged and skips the question, reporting only which port will be used.

`--port` is now a nullable flag value, so an explicit `--port 1880` stays distinguishable from an omitted flag, and range validation moved into the flag type - also lets pflag reject a bad value itself and corrects the help
text, previously advertised as `1-65535` while the guard rejected anything below 1025.

When the chosen port is not the default, the installation directory offered by the next question carries it (`/opt/flowfuse-device-1881`), so agents on different ports are not proposed the same location. It remains a suggestion and any absolute path can be typed instead. Scripted installations driven by `--otc` stay non-interactive throughout and keep the plain default directory.

Additionally, this pull request:
* fixes free port check method - now it checks all available network interfaces instead of loopback only
* adjusts port validation approach
* reorganizes the `utils` package

## Related Issue(s)

Closes #730

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


Steve-Mcl (approved):
Cool! Local testing good on windows,

Image

Steve-Mcl:
one v minor niggle - the line "The FlowFuse Device Agent will use port 1888." is output after user choice, but there is no equivalent "Install Dir: C:\opt\flowfuse-device-1888" output after the directory choice!

PR #747, opened 2026-09-17
fix: Remove .. from assets path

hardillb:
## Description

Ensure only assets from the assets directory are loaded from the assets directory

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


Steve-Mcl (commented):
I think there may be existing flows this will break. 
I would pesonally prefer it we contain it by checking the resolved path.

Something like:

```js

const ASSETS_ROOT = path.join(__dirname, 'assets')

handler: (req, res) => {
    const requested = decodeURIComponent(req.url.split('?')[0].replace(/^\/assets\/?/, ''))
    const assetPath = path.resolve(ASSETS_ROOT, requested)
    const relative = path.relative(ASSETS_ROOT, assetPath)
    if (relative.startsWith('..') || path.isAbsolute(relative)) {
        res.writeHead(404)
        res.end()
        return
    }
    if (!fs.existsSync(assetPath)) {
        res.writeHead(404)
        res.end()
        return
    }
    // ... the rest of it ...
```
(UNTESTED)

Also worth of some regression tests TBH - golden and non-golden path:

hardillb:
@Steve-Mcl Why? The UI only provides access to write the device.yml nothing else that I can see

knolleary:
Checking the url for `..` will miss url encoded path elements that obscure `..`. It is better to validate the resolved path is under the assert dir.

hardillb:
Fine, I have pushed an updated change.

But given the only `asset` is a favicon I'd equally be in favour of just totally removing the whole `assets` route

allthedoll (approved):
Lgtm

knolleary (approved):

PR #749, opened 2026-09-23
feat(installer): Add possibility to re-use existing system-wide Node.js runtime while installing the Device Agent

### LINKED ISSUE #392: Installer should detect existing NodeJS runtime
### Description

The Device Agent installer should detect if there is already NodeJS runtime environment available on the system. Once detecteted, it should compare the versions and ask user for a decision which one should be used for the installation.
If the version is lower than we exepect, installer should print a warning. Also, we should consider installing different version of the Device Agent.

#### Customer Value

Having multiple versions of NodeJS on your machine can get very messy very quickly, causing headaches for customers already with NodeJS installed. The recommended install route with the installer also assumes no NodeJS is present, which in a lot of cases won't be appropriate, this will cover that use case. 

This is also relevant for https://github.com/FlowFuse/device-agent/issues/488 where the Opto 22 devices have limited storage, but already have NodeJS installed, us _not_ installing NodeJS will prevent https://flowfuse.com/docs/device-agent/install/device-agent-installer/#error%3A-disk-space-check-failed occurring 

### Epic/Story

_No response_

ppawlowski:
## Description

This pull request adds the ability to detect and reuse an existing system-wide Node.js runtime to install and run the Device Agent.

Additionally, this pull request adds Microsoft Windows File Properties and Version Info for the Windows executable created via the Makefile. The release pipeline will be updated in a separate PR. This is the quickest and easiest solution to prevent Windows Defender from quarantining the binary when false-positive `Trojan:Win32/Bearfoos.A!ml` is detected.

## Related Issue(s)

Closes https://github.com/FlowFuse/device-agent/issues/392

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


installer/go/pkg/nodejs/system.go
```diff
@@ -0,0 +1,218 @@
...
+//   - nodePath: The absolute path to the candidate node binary
+//
+// Returns:
+//   - string: The reported version, without the leading "v"
+//   - error: An error if the binary cannot be run or reports nothing usable
+func systemNodeVersion(nodePath string) (string, error) {
+	var versionCmd *exec.Cmd
+	switch runtime.GOOS {
+	case "linux", "darwin":
+		versionCmd = exec.Command("sudo", "-n", "-u", utils.ServiceUsername, nodePath, "--version")
```
hardillb: Do we need to run with sudo here just to get the version?

hardillb (commented):
Quick first pass just reading the code all looks good apart from the question about needing `sudo` to just run `node --version` on Linux/OSx

Will try and build it and test it locally on Linux in the morning

installer/go/pkg/nodejs/system.go
ppawlowski: Yes, we execute this command as a user who is about to run the Device Agent as a service.

hardillb (approved):

Steve-Mcl (commented):
Tested this on Windows (elevated cmd, real install). A couple of things might be worth a looksee??


### Non default install locations arent found.
`systemNodeCandidates()` only checks `%ProgramFiles%\nodejs\node.exe` and `%ProgramFiles(x86)%\nodejs\node.exe`. The nodejs.org MSI defaults there, but it lets you pick a custom `INSTALLDIR`. Also, a Chocolatey install to a non-default path won't be found either. Might be worth reading the install location from the registry instead of hardcoding the path - the Node.js MSI registers itself under the Windows uninstall key with an `InstallLocation` value, which would catch a moved install without having to search PATH. Not a total blocker (could be picked up as an iteration) but it is a a gap! I've used devices where C: is a small and apps go on D: or other location & many other variations (i.e. it is not that rare!)

While on this topic, might be worth checking for write filtered volumes too! Some locked down Windows IoT/Embedded boxes run UWF (Unified Write Filter), which discards writes to a protected volume on reboot unless the path is explicitly excluded. If the install directory sits inside a protected volume (likely as it defaults to c:\opt) , the whole install disappears on the next reboot. TBF, this isnt specific to this PR, but might be worth a callout in the README?

### Silent fallback on an old Node.js
I've got Node 20.19.1 at the default `%ProgramFiles%\nodejs` location (I do use `fnm` for later versions), npm is right beside it. The installer's default `--nodejs-version` floor is `22.23.0`, so my install was skipped for being too old and the installer resorted to downloading a bundled copy with no output saying why. That does matche TESTING.md R2 step 2, so it's working as designed, but "found nothing" and "found something too old" look identical. As the Device Agent's own `package.json` only requires `node >=18.x`, this might trip poeple up who beleive they have a usable Node ver installed. Perhaps consider adding a line in the output when users node is is rejected for version? something like `Found Node.js 20.19.1, but 22.23.0 is required - downloading newer runtime`, might save a support question?!?  PS, the README could say this too!

ppawlowski:
@Steve-Mcl 

Node.js installation suggestions addressed in:
* https://github.com/FlowFuse/device-agent/issues/755
* https://github.com/FlowFuse/device-agent/issues/756

Information to a user about a system-wide runtime detected but below the required version surfaced in https://github.com/FlowFuse/device-agent/pull/749/changes/0f9421e127a7e79c36a764c71e6590c11ff8e9f8

Regarding Unified Write Filter - we have never faced this problem; let's skip it until someone actually reports this as an issue.

Steve-Mcl (approved):
Thanks for raising review comments as issues

